PR #8, opened 2025-02-20
Create API.md

owenhalpert:
### Description
Add a readable API document to the root directory describing the contract between the server and client.

### DIFF
--- a/API.md
+++ b/API.md
@@ -1,0 +1,109 @@
+# Remote Vector Service - API Contract
+
+## Endpoints
+
+### Trigger Build
+
+```
+POST /_build
+
+Request Parameters
+{
+    "repository_type": "s3", // Derived from repository settings
+    "container_name": "VectorBucketName", // Derived from repository settings
+    "object_path": "VectorBlobPath", // Directory all vector/doc ID blobs are written to
+    "tenant_id": "UniqueClusterID", // Unique identifier for the cluster making the request
+    "dimension": 768,
+    "doc_count": 1000000,
+    "data_type": "fp32", // {fp16, fp32, byte, binary}
+    "engine": "faiss",
+    "index_parameters": {
+        "space_type": "l2",
+        "algorithm": "hnsw",
+        "algorithm_parameters": {
+            "ef_construction": 100,
+            "m": 16
+        }
+    }
+}
+
+Request Response:
+{
+    "job_id" : "String" // Unique identifier per build
+}
+```
+
+#### Index Parameters
+
+* The only required parameters are `repository_type`, `container_name`, `object_path`, `dimension`, and `doc_count`. The rest will follow k-NN's existing precedent for defaults.
+* The top two parameters are derived by the KNN plugin from the customer's repo setting. `container_name` is used to specifically refer to the "bucket" (or non-S3 equivalent), rather than the name of the repository itself.
+* Tenant ID is included to be used for billing, authorization, etc
+* Dimension, doc count, and data type are specifically placed on the first JSON level so the build service can quickly use them first to calculate workload size
+* Engine: If in the future we have different workers hosting different engines, like Lucene etc. this parameter will act as an extension point. This is not a required parameter.
+* Space type is inside `index_parameters` since it is only used in index creation
+* By including an `algorithm` field in index parameters, we leave the door open for IVF, future algorithms
+* The existing algorithm parameters will be type checked on the client and server side, but the size of the map is variable to allow for future parameter additions
+
+#### Error codes
+
+- `500 Internal Server Error` if there’s an unexpected issue while processing the request. 
+- `507 Out of Memory Exception` if worker does not have memory to handle the request. 
+- `409 Request Conflict` if a conflicting request exists.
+
+### Get Status
+
+```
+GET /_status/{job_id}
+
+Request Response:
+{
+    "task_status" : "String", //RUNNING_INDEX_BUILD, FAILED_INDEX_BUILD, COMPLETED_INDEX_BUILD
+    "index_path" : "String" // Null if not completed
+    "error": "String"
+}
+```
+
+Client can expect an error in “error” if task_status == `FAILED_INDEX_BUILD`.
+
+
+#### Error codes
+- `404 Not Found` if job ID does not exist. 
+
+
+### Authentication
+
+If authentication is configured, the endpoint must support Basic Authentication. The server is expected to return a 401 Unauthorized status code if authentication is enabled and credentials are missing/incorrect. If there is no authentication configured, the server will process requests with no Authentication header.
+
+* * *
+
+## Error Handling
+
+#### Retriable Request Errors
+
+Some HTTP requests may receive a response with an error status code. The client will agree to resend an HTTP request that receives any of the following errors:
+
+| Status Code | Description | Notes |
+|-------------|-------------|--------|
+| 408 | Request Timeout | The server took too long to respond. Retrying may help |
+| 429 | Too Many Requests | Indicates rate-limiting. Retry after the delay specified in the Retry-After header |
+| 500 | Internal Server Error | A server-side issue that may resolve itself. Retrying is optional |
+| 502 | Bad Gateway | Suggests a temporary network issue or service stack disruption that may self-correct |
+| 503 | Service Unavailable | May be due to temporary service outages or in-progress deployments |
+| 504 | Gateway Timeout | A downstream server (e.g., DNS) didn’t respond in time. Retrying may resolve the issue |
+| 509 | Bandwidth Limit Exceeded | Server has exceeded its allocated bandwidth limit |
+
+
+#### Non-Retriable Request Errors
+
+The following failure codes have been designated to signify that the build request should be abandoned. These status codes will throw an exception with a descriptive error, and the node will fall back to CPU build. Most will require intervention (bad auth, bad endpoint, unhealthy build service)
+| Status Code | Description | Notes |
+|-------------|-------------|--------|
+| 400 | Bad Request | The client sent an invalid request. Fix the issue in the request before trying again. |
+| 401 | Unauthorized | Authentication is required or failed. Fix the authentication issue before trying again (e.g. get a fresh token). |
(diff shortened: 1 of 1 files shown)


API.md
```diff
@@ -0,0 +1,114 @@
...
+**404 (Not Found):** The requested resource does not exist. Retrying will not succeed *unless the resource becomes available later* due to some background processing.\
+**405 (Method Not Allowed):** The HTTP method used is not supported. Retrying with the same method will not resolve the issue.\
+**409 (Conflict):** Indicates a conflict in the request such as unique constraints for referential integrity. Retrying without addressing the conflict will continue to fail.\
+**422 (Unprocessable Entity):** The server understands the request but cannot process it due to semantic errors. Fix the issue in the request before trying again.
+
+#### Client Retry Behavior
+
+In our initial implementation, the client will retry one time for a total of two requests. For retriable errors, the client will look for a `Retry-After` header, and if this is not found, the base delay will be 100 milliseconds. For throttling errors (429, 503) the client will  may exponentially backoff before sending subsequent requests, though this only applies to implementations with greater than one retry attempt.
```
rchitale7, line 114: last sentence has a typo

rchitale7 (approved):
LGTM!

API.md
```diff
@@ -0,0 +1,114 @@
...
+## Endpoints
+
+### Trigger Build
+
+```
+POST /_build
+
+Request Parameters
+{
+    "repository_type": "S3", // Derived from repository settings
```
jed326, line 12:
```suggestion
    "repository_type": "s3", // Derived from repository settings
```

API.md
```diff
@@ -0,0 +1,114 @@
...
+
+* * *
+
+## Error Handling
+
+#### Retriable Request Errors
+
+Some HTTP requests may receive a response with an error status code. The client will agree to resend an HTTP request that receives any of the following errors:
+
+>**408 (Request Timeout)**\
```
jed326, line 92: This might look better as a table but not a blocker

jed326 (approved):

API.md
```diff
@@ -0,0 +1,114 @@
...
+**404 (Not Found):** The requested resource does not exist. Retrying will not succeed *unless the resource becomes available later* due to some background processing.\
+**405 (Method Not Allowed):** The HTTP method used is not supported. Retrying with the same method will not resolve the issue.\
+**409 (Conflict):** Indicates a conflict in the request such as unique constraints for referential integrity. Retrying without addressing the conflict will continue to fail.\
+**422 (Unprocessable Entity):** The server understands the request but cannot process it due to semantic errors. Fix the issue in the request before trying again.
+
+#### Client Retry Behavior
+
+In our initial implementation, the client will retry one time for a total of two requests. For retriable errors, the client will look for a `Retry-After` header, and if this is not found, the base delay will be 100 milliseconds. For throttling errors (429, 503) the client may exponentially backoff before sending subsequent requests, though this only applies to implementations with greater than one retry attempt.
```
navneet1v, line 114: why we need to add this here? It should be part of k-NN docs.

API.md
```diff
@@ -0,0 +1,114 @@
...
+
+Some HTTP requests may receive a response with an error status code. The client will agree to resend an HTTP request that receives any of the following errors:
+
+>**408 (Request Timeout)**\
+**429 (Too Many Requests)**\
+**500 (Internal Server Error)**\
+**502 (Bad Gateway):** Suggests a temporary network issue or service stack disruption that may self-correct.\
+**503 (Service Unavailable):** May be due to temporary service outages or in-progress deployments.\
+**504 (Gateway Timeout):** A downstream server (e.g., DNS) didn’t respond in time.\
+**509 (Bandwidth Limit Exceeded)**
```
navneet1v, line 98: this has no details.

API.md
```diff
@@ -0,0 +1,114 @@
...
+- `404 Not Found` if job ID does not exist. 
+
+
+### Authentication
+
+If authentication is configured:
+
+```
+Authentication:
+- Endpoints must support Basic Authentication
```
navneet1v, line 79:
```suggestion
- Endpoints may/maynot support Basic Authentication
```

owenhalpert: The contract intends to convey that, if authentication is configured, the endpoint must support basic authentication. This is a necessity that the client can expect if they configure authentication. If there is no auth configured, none of these statements apply.

API.md
owenhalpert: I only included descriptions for the error codes that might need them (less common or less descriptive of a name). Should I include all of them?

API.md
owenhalpert: I guess you're right, this isn't really a part of the API contract and more how we choose to interact with it.

API.md
```diff
@@ -0,0 +1,109 @@
...
+### Get Status
+
+```
+GET /_status/{job_id}
+
+Request Response:
+{
+    "task_status" : "String", //RUNNING_INDEX_BUILD, FAILED_INDEX_BUILD, COMPLETED_INDEX_BUILD
+    "index_path" : "String" // Null if not completed
+    "error": "String"
```
jed326, line 62: Optional -
```suggestion
    "error_message": "String"
```

jed326 (dismissed):

API.md
```diff
@@ -0,0 +1,109 @@
...
+    "repository_type": "s3", // Derived from repository settings
+    "container_name": "VectorBucketName", // Derived from repository settings
+    "object_path": "VectorBlobPath", // Directory all vector/doc ID blobs are written to
+    "tenant_id": "UniqueClusterID", // Unique identifier for the cluster making the request
+    "dimension": 768,
+    "doc_count": 1000000,
+    "data_type": "fp32", // {fp16, fp32, byte, binary}
+    "engine": "faiss",
+    "index_parameters": {
+        "space_type": "l2",
```
rchitale7, line 21: missed this earlier, but do we need to also specify the other possible space types here?

API.md
owenhalpert: looks good to me, cc @rchitale7

jed326 (approved):

rchitale7 (approved):

API.md
```diff
@@ -0,0 +1,110 @@
...
+    "repository_type": "s3", // Derived from repository settings
+    "container_name": "VectorBucketName", // Derived from repository settings
+    "vector_path": "VectorBlobPath", // File path all vector blobs are written to
+    "doc_id_path" : "DocIdPath", // File path all doc Ids are written to
+    "tenant_id": "UniqueClusterID", // Unique identifier for the cluster making the request
+    "dimension": 768,
+    "doc_count": 1000000,
+    "data_type": "fp32", // {fp16, fp32, byte, binary}
+    "engine": "faiss", // {faiss, nmslib, lucene}
+    "index_parameters": {
```
jmazanec15, line 21: Will this mimic the mapping? I guess one question would be do we want to also potentially allow the index build service the freedom to choose the parameters based on high level signals, like "BEST_RECALL", or "compression_level=32x".

navneet1v: > Will this mimic the mapping?

Ideally it should not mimic the mappings. This interface should evolve of its own.

navneet1v: > I guess one question would be do we want to also potentially allow the index build service the freedom to choose the parameters based on high level signals, like "BEST_RECALL", or "compression_level=32x".

This is a good suggestion. the problem is from k-NN how we can get this info on what customer is looking for.

API.md
```diff
@@ -0,0 +1,110 @@
...
+
+Request Response:
+{
+    "job_id" : "String" // Unique identifier per build
+}
+```
+
+#### Index Parameters
+
+* The only required parameters are `repository_type`, `container_name`, `object_path`, `dimension`, and `doc_count`. The rest will follow k-NN's existing precedent for defaults.
```
rchitale7, line 39: need to change object_path to vector_path and doc_id_path

API.md
jmazanec15: From knn, we can take this information in via mapping parameters: mode, compression_level. For this, we can expand/generalize these parameters.

For this API, I would add a separate field like:
- index_parameter_resolution
- workload_hints
- user_hints

API.md
```diff
@@ -0,0 +1,110 @@
...
+    "container_name": "VectorBucketName", // Derived from repository settings
+    "vector_path": "VectorBlobPath", // File path all vector blobs are written to
+    "doc_id_path" : "DocIdPath", // File path all doc Ids are written to
+    "tenant_id": "UniqueClusterID", // Unique identifier for the cluster making the request
+    "dimension": 768,
+    "doc_count": 1000000,
+    "data_type": "fp32", // {fp16, fp32, byte, binary}
+    "engine": "faiss", // {faiss, nmslib, lucene}
+    "index_parameters": {
+        "space_type": // {l2, cosinesimil, l1, linf, innerproduct, hamming}
```
jmazanec15, line 22: Can we bring this outside of index-parameters? Its more of a data set set defining property as opposed to an index parameter.

owenhalpert: The idea on the split was to have workload defining variables (doc count, data type, etc) immediately available at the first JSON level and keep the index creation parameters inside to be used once we start building the index. So space type was placed inside because it's used in index creation whereas data type for example is more for initially identifying the size of the file.

jmazanec15: I see - then does engine need to be top level?

API.md
owenhalpert: https://github.com/opensearch-project/k-NN/issues/2464#issuecomment-2683206396

owenhalpert:
There are other successful HTTP codes outside of 200 (other 2xx codes). For simplicity can we add to the contract that the server will return 200 in the case of a successful request? Unless you have some detail you want to convey with other accepted codes @rchitale7

navneet1v (approved):

API.md
```diff
@@ -0,0 +1,112 @@
...
+## Endpoints
+
+### Trigger Build
+
+```
+POST /_build
+
+Request Parameters
+{
+    "repository_type": "s3", // Derived from repository settings
```
jed326, line 12: Not going to block this PR, but I think the "comments" should be moved to a table instead, similar to https://opensearch.org/docs/latest/search-plugins/knn/knn-index/#method-definitions for example.

jed326 (approved):

API.md
owenhalpert: Agree

Issue #12, opened 2025-02-25
[Implementation] Remote Vector Index Builder - Create Vectors Dataset and Upload Index Tasks


rchitale7:
### Description
This issue tracks the implementation of the `create_vectors_dataset` and `upload_index` tasks - these tasks handle downloading vectors from a Remote Store repository, and uploading the knn index to the Remote Store repository, respectively. The `build_index` task implementation will be tracked in a separate issue. All 3 of these tasks are utilized by the Remote Vector Index Builder Service to construct a knn index on hardware accelerators (such as GPUs), for an Opensearch cluster. For more background on the Remote Vector Index Builder Service, please read: https://github.com/opensearch-project/k-NN/issues/2293

The tasks are part of the `core` Remote Vector Index Builder Python library. In the initial implementation, only the s3 Remote Store repository will be supported. However, the tasks must be designed to support other repositories (such as Azure Blob Storage, GCP) in the future. 

### Sub-tasks

- [x] Create a draft PR implementation, get feedback from others
- [x] Integrate feedback, add unit tests, and do benchmarking - finalize implementation

PR #13, opened 2025-02-25
Create Vectors Dataset and Upload Index Tasks

rchitale7:
### Description
Draft PR for #12. 

This PR implements the `create_vectors_dataset` and `upload_index` tasks. These two tasks - along with the ```build_index``` task - are used to construct a knn index on hardware accelerators, for an OpenSearch cluster. At a high level, `create_vectors_dataset` downloads vectors from the Remote Store repository, while `upload_index` uploads a knn index to the Remote Store repository. Both tasks utilize an Object Store Client to interact with the Remote Store repository; only the s3 implementation of the Object Store client is covered in this PR. 

Note that the s3 readBlob function uses ```download_fileobj``` to directly read a blob into a buffer in memory. This helps avoid an unnecessary write to disk, as the ``` build_index``` task already expects the vector and doc id blobs as arrays in memory. ``` np.frombuffer ``` is used to convert the buffer into a ```numpy ``` array in place, avoiding the creation of an extra copy. 

Out of scope for this PR:

- Remote Store object checksum validation (will come in a separate PR, but part of the same issue)
- ``` build_index``` task (will come in a separate PR and issue)
- Support for additional repositories beyond s3, such as Azure Blob Storage (will come as a separate PR and issue)

### Checklist

- [x] First pass review
- [x] Unit tests
- [x] Documentation
- [ ] Benchmarking with 30 GB vector dataset


remote-vector-index-builder/core/object_store/s3/s3_object_store.py
```diff
@@ -0,0 +1,171 @@
...
+    def read_blob(self, remote_store_path: str) -> BytesIO:
+
+        # Set up progress callback
+        with self._read_progress_lock:
+            self._read_progress = 0
+
+        def callback(bytes_transferred):
+            with self._read_progress_lock:
+                self._read_progress += bytes_transferred
+                logger.info(f"Downloaded: {self._read_progress:,} bytes")
```
jed326, line 106: Is this callback called just once after upload is completed or after each part upload? Maybe this method would be used for something else in the future but I don't really understand what we're doing right now as it looks like we're just unnecessarily locking during the upload.

remote-vector-index-builder/core/tasks.py
```diff
@@ -0,0 +1,65 @@
...
+def create_vectors_dataset(
+    index_build_params: IndexBuildParameters, object_store_config: Dict[str, Any]
+) -> VectorsDataset:
+    object_store = ObjectStoreFactory.create_object_store(
+        index_build_params, object_store_config
+    )
+
+    # read the blobs into a buffer
+    vectors = object_store.read_blob(index_build_params.vector_path)
+    doc_ids = object_store.read_blob(index_build_params.doc_id_path)
```
jed326, line 38: Nit: I think it's safer to pass the buffer into this method rather than expect it as the return value. As it's implemented now it's not really safe to call `object_store.read_blob` multiple times but it's hard to tell without seeing the implementation of `read_blob`

remote-vector-index-builder/core/tasks.py
```diff
@@ -0,0 +1,65 @@
...
+
+    # create numpy arrays for vectors and doc_ids, from buffer
+    # VectorDataset.parse is also responsible for closing the buffer
```
jed326, line 41: Related to the above comment, I think it's safest if you do the buffer management all in the same place instead of creating it in one place and expecting it to be closed in another. It will also make your failure handling a little more complex to ensure you've closed your buffers, etc.

remote-vector-index-builder/core/tasks.py
```diff
@@ -0,0 +1,65 @@
...
+    index_local_path: str,
+) -> None:
+    object_store = ObjectStoreFactory.create_object_store(
+        index_build_params, object_store_config
+    )
+
+    # index_build_params are already validated beforehand by Pydantic
+    # so we can assume the vector_path can be parsed by splitting at each '.'
+    vector_path_without_extension = index_build_params.vector_path.split(".")[:-1]
+    index_remote_path = ".".join(vector_path_without_extension) + ".knnindex"
```
jed326, line 63: Let's just use the local file name here instead of making it generic. The extension is unique to each engine anyways so we can use that for validation on both sides.

remote-vector-index-builder/core/common/models/index_build_parameters.py
```diff
@@ -0,0 +1,69 @@
...
+
+class Algorithm(str, Enum):
+    HNSW = "hnsw"
+    IVF = "ivf"
+
+
+class Engine(str, Enum):
+    FAISS = "faiss"
+    NMSLIB = "nmslib"
+    LUCENE = "lucene"
```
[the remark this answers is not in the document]
rchitale7: will remove

remote-vector-index-builder/core/common/models/vectors_dataset.py
```diff
@@ -0,0 +1,70 @@
...
+from io import BytesIO
+
+import numpy as np
+from common.exceptions import (UnsupportedVectorsDataTypeError,
+                               VectorsDatasetError)
+from common.models.index_build_parameters import DataType
+
+
+@dataclass
+class VectorsDataset:
```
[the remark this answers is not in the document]
rchitale7: agreed, i'll make sure that it's there next time i have a draft

remote-vector-index-builder/core/object_store/s3/s3_object_store.py
```diff
@@ -0,0 +1,171 @@
...
+        "max_bandwidth": None,
+        "io_chunksize": 256 * 1024,  # 256KB
+        "num_download_attempts": 5,
+        "max_io_queue": 100,
+        "preferred_transfer_client": "auto",
+    }
+
+    # boto3 does not automatically handle retries for these exceptions
+    # so, we need our own custom exception handling for robustness
+    RETRYABLE_BOTO3_EXCEPTIONS = ["InternalError"]
```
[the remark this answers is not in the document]
rchitale7: You're right - I didn't see it being explicitly handled here, under the 'standard retry' section: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/retries.html. But looking closer, InternalError has a 500 error code, so it should be auto retried. 

I know we don't have data for this, but i still think there may be situations where boto3 won't retry a download/upload when it actually will be helpful to do so. What are your thoughts on setting RETRYABLE_BOTO3_EXCEPTIONS to an empty list, but keeping the custom retry logic if/when its necessary to retry an error not handled by boto3?

remote-vector-index-builder/core/object_store/s3/s3_object_store.py
rchitale7: I kept it because i thought it would be helpful for debugging. But can remove since it does impact performance.

remote-vector-index-builder/core/tasks.py
rchitale7: so buffer creation and closing should be done in `create_vectors_dataset` function?

remote-vector-index-builder/core/tasks.py
rchitale7: We still need the created knn index to have the same path as the vector blob up until the extension, right? Based off of the blob naming scheme laid out here: https://github.com/opensearch-project/k-NN/issues/2392

So, essentially, this would become:
`index_remote_path = ".".join(vector_path_without_extension) + "." + index_build_params.engine`

remote-vector-index-builder/core/object_store/s3/s3_object_store.py
jed326: So it's called for each part upload then?

remote-vector-index-builder/core/tasks.py
jed326: > We still need the created knn index to have the same path as the vector blob up until the extension, right?

This would be ideal, but it's not strictly enforced as we will just use whatever path the status request sends back.

remote-vector-index-builder/core/object_store/s3/s3_object_store.py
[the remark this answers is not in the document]
jed326: I don't really see this custom retry logic as p0. My preference would be to add this retry customization logic later as/if needed. If you keep the `RETRYABLE_BOTO3_EXCEPTIONS` as an empty list it might not be possible to test it.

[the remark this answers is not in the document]
jed326: I think generally it's pretty safe to assume the SDK itself should know when to retry and when not to so unless we see a strong specific case where it's needed I would vote against including it.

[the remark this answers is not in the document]
rchitale7: thats fair. We can still test with an empty list, but I'll remove custom retries from this PR.

remote-vector-index-builder/core/object_store/s3/s3_object_store.py
rchitale7: For each chunk, yes. I'll leave this is part of a debug config option

rchitale7:
Currently benchmarking each task with a vector dataset, will post results here once completed

navneet1v:
> This helps avoid an unnecessary write to disk, as the ` build_index` task already expects the vector and doc id blobs as arrays in memory. `np.frombuffer` is used to convert the buffer into a `numpy ` array in place, avoiding the creation of an extra copy.

Have we validated this?

rchitale7:
@navneet1v I'll check this during benchmarking. According to the numpy docs, `np.frombuffer` creates a view into the original object, so it should not use extra memory but I will confirm: 

docs: https://numpy.org/doc/2.1/reference/generated/numpy.frombuffer.html

navneet1v:
> @navneet1v I'll check this during benchmarking. According to the numpy docs, `np.frombuffer` creates a view into the original object, so it should not use extra memory but I will confirm.
> 
> docs: https://numpy.org/doc/2.1/reference/generated/numpy.frombuffer.html

can you test with a small like 1GB file and verify this.

rchitale7:
@navneet1v thanks for the call out. Although `np.frombuffer` creates a view of the object, calling `read` on a `bytesIO` object performs a full copy beforehand. So `np.frombuffer` creates a view for the copy, doubling memory usage. I modified the implementation to call `getbuffer()` on the `bytesIO` object instead, which fixed the issue - `getbuffer()` first creates a view, that can then be passed to `np.frombuffer`. Tested with 307 GB file and verified there was no additional memory usage.

remote_vector_index_builder/core/common/models/vectors_dataset.py
```diff
@@ -0,0 +1,119 @@
...
+
+        Raises:
+            VectorsDatasetError: If there are any errors during parsing or validation.
+        """
+        try:
+            # Create a view into the buffer, to prevent additional allocation of memory
+            vector_view = vectors.getbuffer()
+            np_vectors = np.frombuffer(
+                vector_view, dtype=VectorsDataset.get_numpy_dtype(vector_dtype)
+            )
```
jed326, line 108: if `seek(0)` on the buffer is necessary, then I think you should call it here instead to be safe.

remote_vector_index_builder/core/tasks.py
```diff
@@ -0,0 +1,130 @@
...
+        UnsupportedVectorsDataTypeError: If the index_build_params.data_type is not supported
+        UnsupportedObjectStoreTypeError: If the index_build_params.repository_type is not supported
+
+    """
+    object_store = ObjectStoreFactory.create_object_store(
+        index_build_params, object_store_config
+    )
+
+    object_store.read_blob(index_build_params.vector_path, vector_bytes_buffer)
+    object_store.read_blob(index_build_params.doc_id_path, doc_id_bytes_buffer)
```
jed326, line 84: Probably can take as a future improvement, but I think we should be able to parallelize these downloads as well as the `np.frombuffer` with respect to each other. For example no need to wait for the vector blob download to complete before creating the doc id numpy array

jed326 (commented):
Thanks @rchitale7 , this PR is looking pretty good. Mostly some minor comments and questions remaining from me.

remote_vector_index_builder/core/tasks.py
```diff
@@ -0,0 +1,130 @@
...
+core.tasks
+~~~~~~~~~~~~~~~~~
+
+This module contains the tasks necessary to build an index on GPUs.
+These tasks must be run in the following sequence, for a given build request:
+1. create_vectors_dataset
+2. build_index
+3. upload_index
+
+"""
```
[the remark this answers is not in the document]
rchitale7: Yeah, i will write a 'main.py' that will simply call each of the tasks in sequence. But i wanted to separate the tasks in their own file, if the consumer wants to import them individually and do something different. For example, they may want to stream the VectorDataset from buffer to disk for persistence, and then before calling build_index, stream from disk back to buffer.

remote_vector_index_builder/core/object_store/s3/s3_object_store.py
```diff
@@ -0,0 +1,226 @@
...
+
+        try:
+            self.s3_client.download_fileobj(
+                self.bucket,
+                remote_store_path,
+                bytes_buffer,
+                Config=self.transfer_config,
+                Callback=callback_func,
+            )
+            bytes_buffer.seek(0)
```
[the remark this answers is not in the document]
rchitale7: You're right, its not needed since we're passing the memory view directly to numpy.

remote_vector_index_builder/core/common/models/vectors_dataset.py
rchitale7: per previous comment - will remove this

jed326 (approved):

rchitale7:
Benchmarking results for 28.6 GB vector dataset, on a g5.4xlarge machine: 117 seconds total time (for vector download and conversion to numpy array). 

I also tested with `download_file` and `np.fromfile` instead of `download_fileobj` and `np.from_buffer` for the same dataset, and got 119 seconds total time. `download_file` writes to disk, so in theory it should be slower. But in my test, there was actually little difference between the two methods. It's possible that there's a difference for even larger datasets, but I will need to conduct more tests. For now, I'll merge this PR and revisit doing deeper benchmarking in the future.

PR #16, opened 2025-02-28
Create Dataclasses and Builder for GPU Index Build Config

Rajrahane:
### Description
Implements 
1. Base Interfaces for the IndexBuildService orchestrator and its workflow tasks - GPUIndexBuildService, GPUToCPUIndexConverter, CPUIndexWriter
2. Base Response and Config models for GPU and CPU Index.
3. Concrete Faiss specific data models. Response models for Response for GPU and CPU Index.
4. Faiss Index Build Service build_index method to create a GPU Index, transform to CPU index and write CPU Index to local disc filepath, deletes the indexes after writing to file.
5. Utils to create the params required for GPUIndexConfig from the API IndexBuildParameters.
Screenshot 2025-03-21 at 11 28 31 AM

### Issues Resolved


Rajrahane:
Still a WIP
Here's how you can initialize the config-

def create_index_config(**kwargs) -> GPUIndexBuildConfig:
    builder = IndexConfigBuilder()
    director = IndexConfigDirector(builder)
    return director.construct_config(kwargs)

print(create_index_config(metric='cosinesimil', gpu_config={'graph_build_algo': 'NN_DECENT', 'ivf_pq_build_params': {'n_lists': 1040}}))

remote_vector_index_builder/core/common/models/index_builder/__init__.py
```diff
@@ -0,0 +1,22 @@
+# Copyright OpenSearch Contributors
```
rchitale7, line 1: This is pretty redundant to `core/common/models/__init__.py`. Why do we need both?

remote_vector_index_builder/core/index_builder/index_config_builder.py
```diff
@@ -0,0 +1,73 @@
...
+    GraphBuildAlgo,
+    GPUIndexBuildConfig,
+)
+
+
+class IndexConfigBuilder:
+    def __init__(self):
+        self._hnsw_config: Optional[IndexHNSWCagraConfig] = None
+        self._gpu_config: Optional[GPUIndexCagraConfig] = None
+        self._metric: SpaceType = SpaceType("l2")  # default metric
```
rchitale7, line 17: Do we need to set the default again over here? It already looks like its being set here: https://github.com/opensearch-project/remote-vector-index-builder/pull/16/files#diff-e4fe4cee43117aced2d05b941e98ac8e58f6a5104f617ba89c33d1036e4b0718R26

remote_vector_index_builder/core/common/models/index_builder/ivf_pq_build_cagra_config.py
```diff
@@ -0,0 +1,50 @@
...
+
+    # The bit length of the vector element after compression by PQ.
+    # Possible values: [4, 5, 6, 7, 8].
+    # Hint: the smaller the 'pq_bits', the smaller the index size and the
+    # better the search performance, but the lower the recall.
+    pq_bits: int = 8
+
+    # The dimensionality of the vector after compression by PQ. When zero, an
+    # optimal value is selected using a heuristic.
+    # pq_bits` must be a multiple of 8.
```
rchitale7, line 31: it looks like theres some validation we can do for some of these values. Can we add a validator function to check if pq_bits is a multiple of 8?

remote_vector_index_builder/core/index_builder/index_config_builder.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def set_hnsw_config(self, params: Dict[str, Any]) -> "IndexConfigBuilder":
```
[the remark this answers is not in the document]
Rajrahane: It doesn't recognize the itself as a datatype if we remove the quotes. Just the way it works, hence the quotes

remote_vector_index_builder/core/index_builder/index_config_builder.py
Rajrahane: I set this class variable to a default just to keep it non null

the build method passes this variable as a param to the data class constructor. 
https://github.com/opensearch-project/remote-vector-index-builder/pull/16/files#diff-c4202e71643e2e39cb6a8505a00019cbe2abe7fe43cca55e926a472d310720bf

remote_vector_index_builder/core/common/models/index_builder/__init__.py
Rajrahane: Will delete this file

remote_vector_index_builder/core/common/models/index_builder/ivf_pq_build_cagra_config.py
Rajrahane: Need to discuss this a bit
I referenced the documentation by faiss while writing the comments for all the class variables

https://github.com/facebookresearch/faiss/blob/main/faiss/gpu/GpuIndexCagra.h#L86

In one line it says pq_bits= 8 required, the next line it says, if pq_bits!=8, pq_dim should be multiple of 8

remote_vector_index_builder/core/common/models/index_builder/__init__.py
rchitale7: Can you delete on the next commit :)

remote_vector_index_builder/core/common/models/index_builder/ivf_pq_build_cagra_config.py
rchitale7: Ok, this can be done as a follow up when we start tuning these parameters

remote_vector_index_builder/core/common/models/index_builder/gpu_index_build_config.py
```diff
@@ -0,0 +1,26 @@
...
+from dataclasses import dataclass, field
+
+from ..index_build_parameters import SpaceType
+
+from .gpu_index_cagra_config import GPUIndexCagraConfig
+from .index_hnsw_cagra_config import IndexHNSWCagraConfig
+
+
+@dataclass
+class GPUIndexBuildConfig:
```
rchitale7, line 17: This applies for all classes - can you add a docstring that explains what each class does? Like this: https://github.com/opensearch-project/remote-vector-index-builder/blob/main/remote_vector_index_builder/core/common/models/vectors_dataset.py#L18

remote_vector_index_builder/core/index_builder/faiss_index_builder.py
```diff
@@ -0,0 +1,195 @@
...
+from core.common.models import (
+    GPUIndexBuildConfig,
+    GPUIndexCagraConfig,
+    IndexHNSWCagraConfig,
+    SpaceType,
+)
+from core.index_builder.faiss_index_config_builder import (
+    FaissIndexConfigBuilder,
+)
+import faiss
```
rchitale7, line 17: nit - can we move this import to the top of the file

remote_vector_index_builder/core/index_builder/index_config_builder.py
rchitale7: My point more so was that you're setting the same default in multiple places, which makes the code more confusing. It happens here as well: https://github.com/opensearch-project/remote-vector-index-builder/pull/16/files#diff-e2de99edba10090fcd6ff23547e21f1894181992845ccc0cd59a36e80f429ed6R54 and here: https://github.com/opensearch-project/remote-vector-index-builder/pull/16/files#diff-02070f52dc26d05c382b370ec76f6265cd8ba0f8774a1c24ef3c40fd1290715aR58. Can you consolidate the logic to only set the default in one place?

remote_vector_index_builder/core/common/models/index_builder/gpu_index_cagra_config.py
```diff
@@ -0,0 +1,34 @@
...
+    # Degree of output graph
+    graph_degree: int = 32
+    # ANN Algorithm to build the knn graph
+    graph_build_algo: GraphBuildAlgo = GraphBuildAlgo.IVF_PQ
+
+    store_dataset: bool = False
+    # GPU Device on which the index is resident
+    device: int = 0
+
+    ivf_pq_build_config: IVFPQBuildCagraConfig = field(
```
rchitale7, line 28: Based on how you're building the `GPUIndexCagraConfig` object, I don't think you need these defaults.

remote_vector_index_builder/core/index_builder/models/create_gpu_index_response.py
rchitale7, line 1: can this file be moved to common/models

remote_vector_index_builder/core/index_builder/faiss_index_builder.py
```diff
@@ -0,0 +1,195 @@
...
+                "ivf_pq_build_params": {
+                    "pq_dim": indexBuildParameters.dimension,
+                    "n_lists": calculate_ivf_pq_n_lists(indexBuildParameters.doc_count),
+                }
+            },
+            metric=indexBuildParameters.index_parameters.space_type,
+        )
+
+        # Create a faiis equivalent version of gpu index build config
+        faiss_gpu_index_cagra_config = self._create_faiss_gpu_index_config(
```
rchitale7, line 175: Will this code be extensible to future engines/methods for constructing an index on gpus (like lucene)?

remote_vector_index_builder/core/index_builder/faiss_index_config_builder.py
```diff
@@ -0,0 +1,145 @@
...
+             A configured FAISS IVFPQBuildCagraConfig object with parameters for:
+            - kmeans training set fraction
+            - kmeans iteration count
+            - Product Quantization bits and dimensions
+            - Number of inverted lists (kmeans clusters)
+            - Memory allocation strategy
+        """
+
+        if not self._ivf_pq_build_config:
+            self._ivf_pq_build_config = IVFPQBuildCagraConfig()
```
rchitale7, line 59: Why do we set the default again over here, when we're setting it here: https://github.com/opensearch-project/remote-vector-index-builder/pull/16/files#diff-c4202e71643e2e39cb6a8505a00019cbe2abe7fe43cca55e926a472d310720bfR64

remote_vector_index_builder/core/common/models/index_builder/gpu_index_cagra_config.py
rchitale7: nevermind, i see why you need them now.

rchitale7:
@Rajrahane For general feedback - I feel that the `index_config_builder.py` and `index_config_director.py` classes are unnecessary, and can be consolidated into `faiss_index_builder.py` and `faiss_index_config_builder.py`. It will help make the code more understandable.

remote_vector_index_builder/core/common/models/index_builder/gpu_index_cagra_config.py
rchitale7: Or, you can keep these defaults, and change the way that you're building the `GPUIndexCagraConfig` object.

remote_vector_index_builder/core/index_builder/models/create_gpu_index_response.py
Rajrahane: The scope of the response model is restricted to the index builder. Hence keeping it here.

Rajrahane:
I've tried consolidating creation of the GPUIndexBuildConfig to a simpler one function, rather than the elaborate index_config_builder and index_config_director as suggest by Rohan

The issue is that it was Difficult to parse a complex multilevel dict with 
GPUIndexBuildConfig 
-> GPUIndexCagraConfig ( with IVFPQBuild and IVFPQSearch configs)
-> IndexHNSWConfig

Because the dict sent is complex with missing fields it threw a lot of NULL issues,
 it required me to create the inner objects manually using helpers.
Eg:  ivf_pq_build_config = (
            IVFPQBuildCagraConfig(**ivf_pq_build_params)
            if ivf_pq_build_params
            else IVFPQBuildCagraConfig()
        )
        
And i used the builder pattern in a different filejust to make it more structured, else the single function would be very large.       

Also had to maintain the single responsibility of the main function build_gpu_index of the FaissIndexBuilder.py file to only call all the steps ->
1. build the core dataclass from dict
2. build the faiss binding config version from the dataclass
3. create the gpu index
4. create the cpu index and write to disk

remote_vector_index_builder/core/index_builder/faiss_index_builder.py
Rajrahane: Yes
Here's what i think
we can have an abstract class IndexBuilder with the build_gpu_index method
The FaissIndexBuilder exends implements this class
In the tasks.py, a switcher will instantize the correct IndexBuilder instance, say LuceneIndexBuilder, FaissIndexBuilder and call the build_gpu_index method

For now keeping it simple since we don't have another engine to work on

Rajrahane:
Also the problem to solve in a gist
Goal is to completely configure the GPUIndexConfig from the IndexBuildParameters and convert it into Faiss IndexConfig bindings

before feeding it to the _create_gpu_index and _create_and_write_cpu_index_to_file helper methods

rchitale7:
One other thing to think about - in the future, we may want the remote vector build service to choose the faiss hyperparameters based on 'hints' from the `IndexbuildParameters`. See https://github.com/opensearch-project/remote-vector-index-builder/pull/8#discussion_r1970107536 for more details. 

That's another reason why i feel the current implementation of the `IndexConfigBuilder` and `IndexConfigDirector` classes doesn't make sense, as it will be hard to modify them to work in this scenario. So I agree that keeping the logic to construct the `GPUIndexBuildConfig` object should be in a separate file. But, I don't think we should use a builder pattern to construct this `GPUIndexBuildConfig` object.

remote_vector_index_builder/core/index_builder/faiss_index_builder.py
```diff
@@ -0,0 +1,263 @@
...
+        index_build_config = None
+        faiss_gpu_index_cagra_config = None
+        create_gpu_index_response = None
+
+        try:
+            # Set number of threads for parallel processing
+            faiss.omp_set_num_threads(self.omp_num_threads)
+
+            # Create a structured GPUIndexBuildConfig having defaults, from a partial dictionary set with index params
+            try:
```
jed326, line 203: I don't like the presence of nested try blocks like this, especially I don't see the purpose of the outer one here. Can we be a little more precise with the error handling?

jed326:
I also feel the abstraction here with `IndexConfigBuilder` and `IndexConfigDirector` is a little awkward, for a few reasons:
1. `IndexConfigBuilder` looks like it's specific to HNSW (coupled with CAGRA, despite the generic `_gpu_config` name). That's not necessarily a problem in and of itself, but at the very least we should name it as such
2. `IndexConfigDirector` seems like it only exists because `IndexConfigBuilder` is not a generic abstraction and only supports HNSW / CAGRA
3. Hypothetically let's say we want to support both an additional ANN algorithm from the vector engine as well as a new ANN algorithm on the GPU side. Do we then need `n` x `m` `IndexConfigBuilder` implementations? (where `n` is the number of vector engine algorithms, and `m` is the number of GPU algorithms). I think this is similar to what @rchitale7 is calling out too.
4. I also agree with https://github.com/opensearch-project/remote-vector-index-builder/pull/16#issuecomment-2705230957

Rajrahane:
Here's what i think
For Engine Switching (n engines)
we can have an abstract class IndexBuilder with the build_gpu_index method
The FaissIndexBuilder exends implements this class
In the tasks.py, a switcher will instantize the correct IndexBuilder instance, say LuceneIndexBuilder, FaissIndexBuilder and call the build_gpu_index method

For algorithm switching:  (m supported algorithms for an engine)
The specific Engine class, FaissIndexBuilder can have private methods for the m GPU Algorithms that it calls based on the indexBuildParameter.algorithm parameter

jed326:
I was referring to `n = {"HNSW", "IVF", etc.}` and `m = { 'CAGRA, etc.}`

jed326 (commented):
A couple more high level feedback points:
1. Clearly distinguish between `core.common.model.index_builder` and `core.index_builder` modules. This by itself is making everything here extremely confusing.
2. Clearly distinguish between the data classes require for GPU graph build and the data classes for CPU graph conversion. IMO these should be in separate modules. Please add a doc string to these classes for the same (there's already a comment from @rchitale7 on this)
3. IMO `FaissIndexBuilder` should only be responsible for building the GPU graph (`_create_gpu_index`). The coordination of creating the configs, converting graph to CPU, etc should be a separate component. Please work with @rchitale7 to clearly delineate if this should go into `tasks.py` or somewhere else.
4. The null checking needs to be cleaned up and consolidated into fewer places. This should also naturally clean up your try/catch statements as currently you're just naively try/catching everything
5. Comments should not describe the code, they should explain "why/how", not "what". For example:
```
class FaissIndexBuilder:
    """
    Class exposing the the build_gpu_index method for building a CPU read compatible Faiis GPU Index
    """
```
It is immediately obvious by looking at the code that this class exposes `build_gpu_index`, but more specifically the class is doing much more than that (creating the config, doing GPU graph build, converting it to CPU, etc).

remote_vector_index_builder/core/common/models/index_builder/index_hnsw_cagra_config.py
```diff
@@ -0,0 +1,28 @@
...
+from dataclasses import dataclass
+
+
+@dataclass
+class IndexHNSWCagraConfig:
+    # expansion factor at search time
+    ef_search: int = 256
+
+    # expansion factor at construction time
+    ef_construction: int = 40
```
navneet1v, line 17: Please use 100, 100 here. As those are the defaults of k-NN

remote_vector_index_builder/core/common/models/index_builder/ivf_pq_build_cagra_config.py
```diff
@@ -0,0 +1,50 @@
...
+    # The number of iterations searching for kmeans centers (index building).
+    kmeans_n_iters: int = 10
+    # The fraction of data to use during iterative kmeans building.
+    kmeans_trainset_fraction: float = 0.1
+
+    # The bit length of the vector element after compression by PQ.
+    # Possible values: [4, 5, 6, 7, 8].
+    # Hint: the smaller the 'pq_bits', the smaller the index size and the
+    # better the search performance, but the lower the recall.
+    pq_bits: int = 8
```
navneet1v, line 27: for all these values please find out the defaults from Faiss Cagra and lets use those defaults.

Rajrahane:
Screenshot 2025-03-10 at 3 24 22 PM

Adding the Design for review

rchitale7:
Thanks @Rajrahane, this looks a lot cleaner. Thinking about this more, i think the `build_gpu_index` function you have in `GPUIndexBuildService` actually should be part of the `GPUIndexConfig` dataclass. Same applies for the `GPUToCPUIndexConverter` and `CPUIndexConfig` classes. Because i think the methods in which the GPU and CPU indices are built are very dependent on the config.

I'm also not sure if we need a separate `CPUIndexWriter` interface, but thats a smaller concern.

jed326:
Thanks @Rajrahane , the class diagram looks reasonable to me. Let's just try to avoid dumping everything into the same module to keep things organized.

Rajrahane:
Screenshot 2025-03-12 at 10 25 56 AM

remote_vector_index_builder/core/common/models/index_builder/faiss/faiss_cpu_index_response.py
```diff
@@ -0,0 +1,43 @@
...
+        Destructor to clean up FAISS resources.
+        Ensures CPU memory is properly freed when the object is destroyed.
+
+        The method handles cleanup by
+        explicitly deleting cpu index to free CPU and system memory
+        """
+
+        try:
+            if self.cpu_index:
+                self.cpu_index.own_fields = True
```
rchitale7, line 40: In the POC this happens before calling `copyTo` to copy the index to CPU format : https://github.com/navneet1v/VectorSearchForge/blob/main/remote-index-build-service/worker/index_builder/gpu/create_gpu_index.py#L100. Since the destructor gets called after the index gets copied to CPU format, setting `own_fields = True` happens after the index is copied to CPU format . Have we tested if its ok to do this?

remote_vector_index_builder/core/common/models/index_builder/faiss/ivf_pq_build_cagra_config.py
```diff
@@ -0,0 +1,140 @@
...
+        if "pq_dim" in params:
+            pq_dim = params["pq_dim"]
+            if pq_dim < 0:
+                raise ValueError(
+                    "IVFPQBuildCagraConfig param: pq_dim must be non-negative"
+                )
+
+            # Check pq_dim constraints based on pq_bits
+            pq_bits = params.get("pq_bits", 8)  # Use 8 if not specified
+            if pq_bits != 8 and pq_dim % 8 != 0:
```
rchitale7, line 96: nice validation check here!

Rajrahane:
Manually Tested the code

Output:
1.9.0
Vector Creation time:  1.8693499565124512
using ivf_pq::index_params nrows 76800, dim 1000, n_lits 277, pq_dim 256
[I] [18:17:22.213133] optimizing graph
[I] [18:17:23.119892] Graph optimized, creating index
Index Build Service time:  14.906126499176025
Upload Index time:  2.1730029582977295

remote_vector_index_builder/core/index_builder/faiss/faiss_index_build_service.py
```diff
@@ -0,0 +1,256 @@
...
+        Returns:
+        CreateGPUIndexResponse: A data model containing the created GPU Index and ID Map Index
+        """
+        faiss_gpu_index = None
+        faiss_id_map_index = None
+        faiss_gpu_index_config = None
+
+        # Create a faiis equivalent version of gpu index build config
+        try:
+            faiss_gpu_index_config = cast(
```
rchitale7, line 75: This is what i was getting at in my earlier comment: https://github.com/opensearch-project/remote-vector-index-builder/pull/16#issuecomment-2712169565  - we can avoid doing this casting if we move the functionality in `build_gpu_index` to the `GPUIndexConfig` dataclass itself, and fully leverage polymorphism.

rchitale7: @jed326 I'm on the fence if this is a change that is necessary to be made. It'll make testing easier, and will remove the need for these try-catch blocks. But the code here technically works. What do you think?

remote_vector_index_builder/core/index_builder/faiss/faiss_index_build_service.py
```diff
@@ -0,0 +1,256 @@
...
+                gpu_index_cagra=faiss_gpu_index, index_id_map=faiss_id_map_index
+            )
+        except Exception as e:
+            if faiss_gpu_index is not None:
+                del faiss_gpu_index
+            if faiss_id_map_index is not None:
+                del faiss_id_map_index
+            raise Exception(f"Failed to create faiss GPU index: {str(e)}") from e
+
+    def convert_gpu_to_cpu_index(
```
rchitale7, line 107: Same thing https://github.com/opensearch-project/remote-vector-index-builder/pull/16#issuecomment-2712169565 applies here

remote_vector_index_builder/core/common/models/index_builder/faiss/faiss_cpu_index_response.py
```diff
@@ -0,0 +1,43 @@
...
+import faiss
+from dataclasses import dataclass
+
+from core.common.models import CPUIndexResponse
+
+
+@dataclass
+class FaissCPUIndexResponse(CPUIndexResponse):
+    """
+    A data class that holds FAISS CPU vector search index components for vector similarity search.
```
jed326, line 17: This class isn't really for `search` though right, we're just using it in the context of building a graph this doesn't land in any search path (at least for now).

remote_vector_index_builder/core/common/models/index_builder/faiss/faiss_cpu_index_response.py
```diff
@@ -0,0 +1,43 @@
...
+            and the cpu index
+    """
+
+    cpu_index: faiss.IndexHNSWCagra
+    index_id_map: faiss.Index
+
+    def __del__(self):
+        """
+        Destructor to clean up FAISS resources.
+        Ensures CPU memory is properly freed when the object is destroyed.
```
jed326, line 32: Are we talking about RAM or actually CPU memory?

remote_vector_index_builder/core/index_builder/__init__.py
jed326, line 1: Think you need the license here still

remote_vector_index_builder/core/index_builder/faiss/faiss_index_build_service.py
jed326: > This is what i was getting at in my earlier comment: [#16 (comment)](https://github.com/opensearch-project/remote-vector-index-builder/pull/16#issuecomment-2712169565) - we can avoid doing this casting if we move the functionality in `build_gpu_index` to the `GPUIndexConfig` dataclass itself, and fully leverage polymorphism.

Agree on this, or we can just have a `to_config()` class method in `GPUIndexConfig` which is implemented in `GPUIndexCagraConfig`. Having the cast here looks to me like we are doing something wrong with the layers of abstraction, what is even the point of `GPUIndexConfig` if we are doing this?

jed326 (commented):
I think I may have misunderstood the design in https://github.com/opensearch-project/remote-vector-index-builder/pull/16#issuecomment-2718621782, but generally speaking I really do not like the empty base classes as that means nothing is actually being abstracted and you need to do casting to even use the class.

remote_vector_index_builder/core/common/models/index_builder/faiss/faiss_cpu_index_response.py
Rajrahane: the 'own_fields' parameter is used during the copyTo function call to delete an existing internal index, if any before copying to CPU index: https://github.com/facebookresearch/faiss/blob/main/faiss/gpu/GpuIndexCagra.cu#L220

When initializing IndexHNSWCagra, the index is set to null
https://github.com/facebookresearch/faiss/blob/main/faiss/IndexHNSW.h#L35

The 'own_fields' is also used during destruction of IndexHNSWCagra like: https://github.com/facebookresearch/faiss/blob/main/faiss/IndexHNSW.cpp#L218

Hence its only actual use for us is during destruction

remote_vector_index_builder/core/common/models/index_builder/index_hnsw_cagra_config.py
Rajrahane: updated

remote_vector_index_builder/core/common/models/index_builder/ivf_pq_search_cagra_config.py
```diff
@@ -0,0 +1,14 @@
...
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+
+from dataclasses import dataclass
+
+
+@dataclass
+class IVFPQSearchCagraConfig:
+    # The number of clusters to search.
+    n_probes: int = 30
```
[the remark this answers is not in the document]
Rajrahane: updated

remote_vector_index_builder/core/common/models/index_builder/ivf_pq_build_cagra_config.py
Rajrahane: updated

remote_vector_index_builder/core/common/models/index_builder/ivf_pq_build_cagra_config.py
Rajrahane: Added validation

remote_vector_index_builder/core/common/models/index_builder/response/faiss_index_id_map.py
```diff
@@ -0,0 +1,40 @@
...
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+
+import faiss
+from dataclasses import dataclass
+
+
+@dataclass
+class FaissIndexIDMap:
```
navneet1v, line 13: I didn't think name of the class will this specific. Can we name it BuildIndexOutput.

remote_vector_index_builder/core/common/models/index_builder/faiss/ivf_pq_build_cagra_config.py
```diff
@@ -0,0 +1,140 @@
...
+                raise ValueError(
+                    "IVFPQBuildCagraConfig param: kmeans_n_iters must be positive"
+                )
+        if "kmeans_trainset_fraction" in params:
+            if not 0 < params["kmeans_trainset_fraction"] <= 1:
+                raise ValueError(
+                    "IVFPQBuildCagraConfig param: kmeans_trainset_fraction must be between 0 and 1"
+                )
+        if "pq_bits" in params:
+            if params["pq_bits"] not in [4, 5, 6, 7, 8]:
```
navneet1v, line 83: for these validation add the code reference of Faiss/Cuvs where these values are provided.

remote_vector_index_builder/core/common/models/index_builder/faiss/ivf_pq_build_cagra_config.py
```diff
@@ -0,0 +1,140 @@
...
+    # individual clusters
+    # This allows to amortize the cost of memory allocation and
+    # reduce the number of data copies during repeated calls to `extend`
+    # (extending the database).
+    #
+    # The alternative is the conservative allocation behavior; when enabled,
+    # the algorithm always allocates the minimum amount of memory required to
+    # store the given number of records. Set this flag to `true` if you prefer
+    # to use as little GPU memory for the database as possible.
+    conservative_memory_allocation: bool = False
```
navneet1v, line 54: this should be true for RemoteIndexBuild service.

remote_vector_index_builder/core/common/models/index_builder/faiss/ivf_pq_build_cagra_config.py
```diff
@@ -0,0 +1,140 @@
...
+            - Memory allocation strategy
+        """
+
+        config = faiss.IVFPQBuildCagraConfig()
+        config.kmeans_trainset_fraction = self.kmeans_trainset_fraction
+        config.kmeans_n_iters = self.kmeans_n_iters
+        config.pq_bits = self.pq_bits
+        config.pq_dim = self.pq_dim
+        config.n_lists = self.n_lists
+        config.conservative_memory_allocation = self.conservative_memory_allocation
```
navneet1v, line 121: there were other parameters which we need to set for cagra index. why I am not seeing those parameters. Parameter include spaceType, store_dataset.

remote_vector_index_builder/core/common/models/index_builder/faiss/faiss_gpu_index_cagra_builder.py
```diff
@@ -0,0 +1,214 @@
...
+    """
+
+    # Degree of input graph for pruning
+    intermediate_graph_degree: int = 128
+    # Degree of output graph
+    graph_degree: int = 64
+    # ANN Algorithm to build the knn graph
+    graph_build_algo: CagraGraphBuildAlgo = CagraGraphBuildAlgo.IVF_PQ
+
+    store_dataset: bool = True
```
navneet1v, line 37: this should be false for us.

remote_vector_index_builder/core/common/models/index_builder/faiss/faiss_gpu_index_cagra_builder.py
```diff
@@ -0,0 +1,214 @@
...
+class FaissGPUIndexCagraBuilder(FaissGPUIndexBuilder):
+    """
+    Configuration class for Faiss GPU Index Cagra
+    Also exposes a method to build the gpu index from the CAGRA configuration
+    """
+
+    # Degree of input graph for pruning
+    intermediate_graph_degree: int = 128
+    # Degree of output graph
+    graph_degree: int = 64
```
navneet1v, line 33: values should be 64 and 32 here which default to m = 16

remote_vector_index_builder/core/common/models/index_builder/faiss/faiss_gpu_index_cagra_builder.py
```diff
@@ -0,0 +1,214 @@
...
+
+
+class FaissGPUIndexCagraBuilder(FaissGPUIndexBuilder):
+    """
+    Configuration class for Faiss GPU Index Cagra
+    Also exposes a method to build the gpu index from the CAGRA configuration
+    """
+
+    # Degree of input graph for pruning
+    intermediate_graph_degree: int = 128
```
navneet1v, line 31: there is another parameter called as refinement, whose default value is 2. Lets use that  parameter too here. Please check this reference: https://github1s.com/facebookresearch/faiss/blob/main/faiss/gpu/GpuIndexCagra.h#L178-L179

test_remote_vector_index_builder/requirements.txt
```diff
@@ -2,4 +2,5 @@ black>=25.1.0,<26.0.0
 flake8>=7.1.0,<8.0.0
 isort>=6.0.0,<7.0.0
 mypy>=1.15.0,<2.0.0
-pytest>=8.0.0,<9.0.0
\ No newline at end of file
+pytest>=8.0.0,<9.0.0
+faiss-cpu==1.10.0
```
navneet1v, line 6: why we have this dependency?

Rajrahane: Added it as a test dependency for faiss resolution

https://github.com/opensearch-project/remote-vector-index-builder/pull/16/files#diff-216f630a295f1e0afb1340288e6d79681b771513001cd01f64ab2614f8229594R20

datamodel FaissBuildIndexOutput faiss.IndexIDMap variable
test fails stating faiss module import error

rchitale7: can you use `sys.modules` to mock the `faiss` import? That is a better approach than adding a new dependency

Rajrahane:
Screenshot 2025-03-21 at 11 28 31 AM

Updated Design

test_remote_vector_index_builder/requirements.txt
Rajrahane: Resolved.
added FaissMock

remote_vector_index_builder/core/index_builder/faiss/faiss_index_build_service.py
```diff
@@ -0,0 +1,256 @@
...
+        cpu_index_config = None
+        gpu_index_response = None
+
+        try:
+            # Set number of threads for parallel processing
+            faiss.omp_set_num_threads(self.omp_num_threads)
+
+            # Step 1a: Create a structured GPUIndexConfig having defaults,
+            # from a partial dictionary set from index build params
+            gpu_index_config_params = {
```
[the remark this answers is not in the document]
Rajrahane: Planned to extract this when a new algo implementation is added. Keeping it inline for now

remote_vector_index_builder/core/index_builder/faiss/faiss_index_build_service.py
Rajrahane: Resolved in redesign

remote_vector_index_builder/core/index_builder/faiss/faiss_index_build_service.py
Rajrahane: Resolved in redesign

remote_vector_index_builder/core/index_builder/faiss_index_config_builder.py
Rajrahane: Resolved in redesign

remote_vector_index_builder/core/index_builder/__init__.py
rchitale7, line 1: need the opensource license here

remote_vector_index_builder/core/common/models/index_builder/faiss/faiss_gpu_index_type
rchitale7, line 1: is this file being used?

Rajrahane: The enum is not actively used in a switch/ if-else, will be needed for the fp16 version being added by Navneet
Added as part of the design we approved

Rajrahane: Added a class documentation

remote_vector_index_builder/core/index_builder/__init__.py
Rajrahane: done

rchitale7 (approved):

remote_vector_index_builder/core/common/models/index_builder/faiss/faiss_gpu_index_type
```diff
@@ -0,0 +1,16 @@
...
+
+from enum import Enum
+
+
+class FaissGPUIndexType(Enum):
+    """
+    Use this enum to switch the type of FAISS GPU Index to build
+    in the FaissIndexBuildService
+    """
+    CAGRA = "CAGRA"
```
jed326, line 16: Is this still being used? Also shouldn't it be a `.py` file?

Rajrahane: The enum is not actively used in a switch/ if-else, will be needed for the fp16 version being added by Navneet
Added as part of the design we approved
Added a class documentation

remote_vector_index_builder/core/common/models/index_builder/faiss/faiss_gpu_index_cagra_builder.py
```diff
@@ -0,0 +1,224 @@
...
+            ivf_pq_build_config=ivf_pq_build_config,
+            ivf_pq_search_config=ivf_pq_search_config,
+        )
+
+    def build_gpu_index(
+        self,
+        vectorsDataset: VectorsDataset,
+        dataset_dimension: int,
+        space_type: SpaceType,
+    ) -> FaissGpuBuildIndexOutput:
```
jed326, line 175: This is more of a nit, but it feels like these classes don't belong under `common/models` as this class is actually doing the GPU graph build rather than being a "model" class

remote_vector_index_builder/core/common/models/index_builder/faiss/faiss_gpu_index_type.py
```diff
@@ -0,0 +1,17 @@
...
+# SPDX-License-Identifier: Apache-2.0
+#
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+
+from enum import Enum
+
+
+class FaissGPUIndexType(Enum):
```
jed326, line 11: I still don't see where this one is being used but I could be missing it, could you point it out for me?

remote_vector_index_builder/core/common/models/index_builder/faiss/faiss_index_hnsw_cagra_builder.py
```diff
@@ -0,0 +1,132 @@
...
+    FaissCPUIndexBuilder,
+)
+
+
+@dataclass
+class FaissIndexHNSWCagraBuilder(FaissCPUIndexBuilder):
+    """Configuration class for HNSW Cagra CPU Index"""
+
+    # expansion factor at search time
+    ef_search: int = 100
```
jed326, line 23: Also a future improvement, but I'm thinking it might be better to keep all of these graph tuning related parameters in a single centralized place (`faiss_graph_parameters` maybe) that includes both the GPU and CPU construction parameters. That way we don't have to dig through the code to see where the values are defined.

jed326 (commented):
Thanks @Rajrahane , this PR mostly LGTM, just a few spots of discussion/nits

remote_vector_index_builder/core/common/models/index_builder/faiss/faiss_gpu_index_type.py
Rajrahane: The enum is not actively used anywhere, in a switch/ if-else, will be needed for the fp16 version being added by Navneet
Added as part of the design we approved for future use.

When a new Faiss index Build type is added, a switcher can be added to pick the correct FaissServiceImpl
And this:
https://github.com/opensearch-project/remote-vector-index-builder/pull/16#pullrequestreview-2683105121
can be done too

Rajrahane: Deleting the file for now after group discussion

remote_vector_index_builder/core/common/models/index_builder/faiss/faiss_gpu_index_cagra_builder.py
Rajrahane: One solution i came up with:

remote-vector-index-builder/
├── remote_vector_index_builder/
│   ├── core/
│   │   ├── services/
│   │   │   └── faiss/
│   │   │       ├── index_builders/
│   │   │       │   ├── __init__.py
│   │   │       │   ├── gpu_index_builder.py     # Abstract GPU builder
│   │   │       │   ├── cpu_index_builder.py     # Abstract CPU builder
│   │   │       │   ├── cagra/                   # Concrete CAGRA implementations
│   │   │       │   │   ├── __init__.py
│   │   │       │   │   ├── faiss_gpu_index_cagra_builder.py
│   │   │       │   │   └── faiss_index_hnsw_cagra_builder.py
│   │   │       │   └── fp16/                    # Concrete FP16 implementations
│   │   │       │       ├── __init__.py
│   │   │       │       └── faiss_gpu_index_fp16_builder.py
│   │   │       └── service.py

with
class FaissGPUIndexCagraBuilder(FaissGPUIndexBuilder):
    """
    Builder class for Faiss GPU Index Cagra
    """
    def __init__(self, config: FaissGPUIndexCagraConfig):
        super().__init__()
        self.config = config
OR
class FaissGPUIndexFP16Builder(FaissGPUIndexBuilder):
    def __init__(self, config: FaissGPUIndexFP16Config):
        super().__init__()
        self.config = config

remote_vector_index_builder/core/common/models/index_builder/faiss/faiss_index_hnsw_cagra_builder.py
Rajrahane: index_build_parameters.py AlgorithmParameters class
model_config = ConfigDict(extra="allow"): to allow extra attributes using the ConfigDict class

PR #17, opened 2025-03-03
Added checksum support to s3 operations. 

rchitale7:
### Description
Adding support for checksums on s3 file download and upload operations. Defaulting to SHA256 for uploads. 
Also, modified the enums for `data_type` and `space_type` in the `IndexBuildParameters` to match what's here: https://github.com/opensearch-project/remote-vector-index-builder/blob/main/API.md

### Issues Resolved
Closes https://github.com/opensearch-project/remote-vector-index-builder/issues/12

### DIFF
--- a/remote_vector_index_builder/core/common/models/index_build_parameters.py
+++ b/remote_vector_index_builder/core/common/models/index_build_parameters.py
@@ -18,16 +18,10 @@ class DataType(str, Enum):
     """Supported data types for vector values.
 
     Attributes:
-        FLOAT32: 32-bit floating point values
-        FLOAT16: 16-bit floating point values
-        BYTE: 8-bit integer values
-        BINARY: Binary data format
+        FLOAT: 32-bit floating point values
     """
 
-    FLOAT32 = "fp32"
-    FLOAT16 = "fp16"
-    BYTE = "byte"
-    BINARY = "binary"
+    FLOAT = "float"
 
 
 class SpaceType(str, Enum):
@@ -35,19 +29,11 @@ class SpaceType(str, Enum):
 
     Attributes:
         L2: Euclidean distance
-        COSINESIMIL: Cosine similarity
-        L1: Manhattan distance
-        LINF: Chebyshev distance
         INNERPRODUCT: Dot product similarity
-        HAMMING: Hamming distance for binary vectors
     """
 
     L2 = "l2"
-    COSINESIMIL = "cosinesimil"
-    L1 = "l1"
-    LINF = "linf"
     INNERPRODUCT = "innerproduct"
-    HAMMING = "hamming"
 
 
 class Algorithm(str, Enum):
@@ -145,7 +131,7 @@ class IndexBuildParameters(BaseModel):
     tenant_id: str = ""
     dimension: int = Field(gt=0)
     doc_count: int = Field(gt=0)
-    data_type: DataType = DataType.FLOAT32
+    data_type: DataType = DataType.FLOAT
     engine: Engine = Engine.FAISS
     index_parameters: IndexParameters = Field(default_factory=IndexParameters)
     model_config = ConfigDict(extra="forbid")
--- a/remote_vector_index_builder/core/object_store/s3/s3_object_store.py
+++ b/remote_vector_index_builder/core/object_store/s3/s3_object_store.py
@@ -47,6 +47,10 @@ class S3ObjectStore(ObjectStore):
     Attributes:
         DEFAULT_TRANSFER_CONFIG (dict): Default configuration for S3 file transfers,
             including chunk sizes, concurrency, and retry settings
+        DEFAULT_DOWNLOAD_ARGS (dict): Default boto3 ALLOWED_DOWNLOAD_ARGS values.
+            Includes encryption and checksum settings.
+        DEFAULT_UPLOAD_ARGS (dict): Default boto3 ALLOWED_UPLOAD_ARGS values.
+            Includes encryption and checksum settings
 
     Args:
         index_build_params (IndexBuildParameters): Parameters for the index building process
@@ -56,7 +60,7 @@ class S3ObjectStore(ObjectStore):
     DEFAULT_TRANSFER_CONFIG = {
         "multipart_chunksize": 10 * 1024 * 1024,  # 10MB
         "max_concurrency": (os.cpu_count() or 2)
-        // 2,  # os.cpu_count can None, according to mypy. If it is none, then default to 1 thread
+        // 2,  # os.cpu_count can be None, according to mypy. If it is none, then default to 1 thread
         "multipart_threshold": 10 * 1024 * 1024,  # 10MB
         "use_threads": True,
         "max_bandwidth": None,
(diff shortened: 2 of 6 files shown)


remote_vector_index_builder/core/common/models/index_build_parameters.py
```diff
@@ -18,36 +18,22 @@ class DataType(str, Enum):
...
-        BYTE: 8-bit integer values
-        BINARY: Binary data format
+        FLOAT: 32-bit floating point values
     """
 
-    FLOAT32 = "fp32"
-    FLOAT16 = "fp16"
-    BYTE = "byte"
-    BINARY = "binary"
+    FLOAT = "float"
```
jed326, line 24: If integrity checking and data type changes are not related to each other then I would prefer to keep these in separate PRs. I'll review them together for now, but if there is a lot of discussion on either topic let's separate them out.

remote_vector_index_builder/core/object_store/s3/s3_object_store.py
```diff
@@ -66,6 +70,50 @@ class S3ObjectStore(ObjectStore):
...
+        "SSECustomerKey": None,
+        "SSECustomerKeyMD5": None,
+        "RequestPayer": None,
+        "ExpectedBucketOwner": None,
+    }
+
+    DEFAULT_UPLOAD_ARGS = {
+        "ACL": None,
+        "CacheControl": None,
+        "ChecksumAlgorithm": "SHA256",
```
jed326, line 86: OpenSearch repository-s3 uses CRC32: https://github.com/opensearch-project/OpenSearch/blob/c328c18f69c08b21aa8a76af270b04a70c5a8069/plugins/repository-s3/src/main/java/org/opensearch/repositories/s3/async/AsyncTransferManager.java#L163

remote_vector_index_builder/core/object_store/s3/s3_object_store.py
```diff
@@ -90,7 +138,26 @@ def __init__(
...
+        self.download_args = S3ObjectStore._create_custom_config(
+            download_args, self.DEFAULT_DOWNLOAD_ARGS
+        )
+
+        upload_args = object_store_config.get("upload_args", {})
+        # Create upload args with validated parameters
+        # This is passed as the 'ExtraArgs' parameter to the boto3 upload API call
+        self.upload_args = S3ObjectStore._create_custom_config(
+            upload_args, self.DEFAULT_UPLOAD_ARGS
+        )
```
jed326, line 160: This is a bit of an "unknown unknown", I would think upload/download args would be configured as a part of the `read/write_blob` methods as these are subject to change on each request. I know right now there's no way to dynamically configure any of these, but if there were then this would not work here.

remote_vector_index_builder/core/object_store/s3/s3_object_store.py
```diff
@@ -101,33 +168,37 @@ def __init__(
...
 
         Returns:
-            TransferConfig: Configured transfer configuration object
+            dict: A merged configuration dictionary suitable for boto3 TransferConfig
+                or ExtraArgs parameters, where custom values take precedence over defaults
         """
-        # Start with default values
-        config_params = self.DEFAULT_TRANSFER_CONFIG.copy()
 
+        # Start with default values
```
jed326, line 189: Doesn't S3 provide some type of integrity checking by default already? https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity.html

remote_vector_index_builder/core/object_store/s3/s3_object_store.py
```diff
@@ -66,6 +70,50 @@ class S3ObjectStore(ObjectStore):
...
 
+    DEFAULT_DOWNLOAD_ARGS = {
+        "ChecksumMode": "ENABLED",
+        "VersionId": None,
+        "SSECustomerAlgorithm": None,
+        "SSECustomerKey": None,
+        "SSECustomerKeyMD5": None,
+        "RequestPayer": None,
+        "ExpectedBucketOwner": None,
+    }
```
jed326, line 81: I don't like the idea of maintaining a copy of these allowed args in our code which we would need to manually keep updated. Can we either 
1. Directly reference what boto3 is keeping for validation (for example, https://github.com/boto/s3transfer/blob/develop/s3transfer/manager.py#L162-L223)
2. Just let boto3 do the validation and if there are invalid arguments the request fails. 

I personally lean towards [2], I don't directly see the benefit of doing argument validation ourselves since boto3 will do it again.

remote_vector_index_builder/core/object_store/s3/s3_object_store.py
rchitale7: in that case, wouldn't the same apply for the transfer config? I can change the `read_blob` and `write_blob` interfaces to include a dictionary parameter that specifies download/upload specific args - what do you think?

remote_vector_index_builder/core/object_store/s3/s3_object_store.py
rchitale7: The idea behind this was to have our own default values. Currently, the only default value we need is checksum mode. If arg validation is not necessary, i can remove all of the fields that are currently set to 'None' - but we still need some way of keeping defaults.

jed326: Removing all the fields that are set to `None` is good, as their presence only serves as a check that the value is valid, which we can leave to boto3 to do.

remote_vector_index_builder/core/object_store/s3/s3_object_store.py
jed326: In my head I was thinking of transfer config as the attributes of the s3 bucket (hence config rather than args), while download/upload args is arguments specific to a download/upload. I know we pass the transfer config to each upload/download call though so that characterization isn't completely accurate. I think for now since there are no plans to make these values dynamically configurable at runtime, we can just leave it as is.

remote_vector_index_builder/core/object_store/s3/s3_object_store.py
rchitale7: It looks like it does - but it does not use CRC32 by default, according to those docs. I still think it makes sense to give the caller of this library some control over these parameters, as they may need to be tuned in the future. In fact, to match OpenSearch's behavior, we should be overriding the s3 integrity defaults with CRC32

jed326: This comment was in regards to the `ChecksumMode` on the download path. Just checking if we actually needed it.

rchitale7: That's a good question. I'm actually not sure - I looked into the S3Transfer code and wasn't able to determine if its enabled by default. My guess is that it is, but to be safe, we should enable it.

remote_vector_index_builder/core/object_store/s3/s3_object_store.py
```diff
@@ -167,14 +192,19 @@ def callback(bytes_transferred):
...
-                Config=self.transfer_config,
+                Config=s3_transfer_config,
                 Callback=callback_func,
+                ExtraArgs=self.download_args,
             )
             return
+        except TypeError as e:
+            raise BlobError(f"Error calling boto3.download_fileobj: {e}") from e
         except ClientError as e:
             raise BlobError(f"Error downloading file: {e}") from e
```
jed326, line 209: Do we need additional handling for checksum errors? Also is there any way to actually test that the checksum is working?

remote_vector_index_builder/core/object_store/s3/s3_object_store.py
navneet1v: So, for default configuration are we going to use CRC checksum or anything else?

remote_vector_index_builder/core/object_store/s3/s3_object_store.py
rchitale7: Checksum errors get raised as `ClientError`, so I don't think it makes sense to add additional error handling. I think testing if the checksum works or not belongs in the integration tests, not unit tests. But I'd need to think more about how to do that.

remote_vector_index_builder/core/object_store/s3/s3_object_store.py
rchitale7: CRC32 checksum for uploads. For downloads, the checksum will be based on the checksum algorithm used when the object was first uploaded to s3.

remote_vector_index_builder/core/object_store/s3/s3_object_store.py
jed326: Can we track the integration tests we need to add later in a GH issue then?

rchitale7: Created issue: https://github.com/opensearch-project/remote-vector-index-builder/issues/19

navneet1v:
@rchitale7 can you please create a gh issue which tracks the performance benchmarks for uploads and downloads for this piece of code?

navneet1v (approved):

rchitale7:
Opened the issue to track benchmarking: https://github.com/opensearch-project/remote-vector-index-builder/issues/23. Will merge this code.

Issue #23, opened 2025-03-05
[Benchmarking] Benchmarking Vector Blob Download and Index Upload


rchitale7:
This issue tracks performance benchmarks for the Remote Vector Index Builder Service vector blob download and index upload. These benchmarks are specific to s3, because that is the only repository currently supported. However, in the future, there will be separate issues tracking performance for GCP/Azure. 

### Background

The Remote Vector Index Builder Service uses `boto3` to handle file downloads and uploads. `boto3` uses a default file transfer configuration, handled by the `s3.transfer.TransferConfig` object. There are many individual parameters in the file transfer configuration - see the [documentation](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/customizations/s3.html#boto3.s3.transfer.TransferConfig) for the full list. Based on manual testing and research, I determined that there are two parameters which greatly impact vector blob download/index upload performance:

- `multipart_chunksize`
    - This is size of each part for multipart transfer requests, for parallelized downloads/uploads
    - If the file is smaller than the chunk size, `boto3` does a single part transfer. This results in sequential download/upload, which is slower. 
- `max_concurrency` 
    - This is the number of threads to use for making a multi part transfer request
    - Increasing the number of threads can result in better performance, due to increased parallelism. However, using more threads may also saturate the connection pool, causing errors and retries. More errors and retries leads to worse performance. 

To test the impact of these two parameters, I tried downloading a 293 MB vector dataset with varying thread counts and multipart chunk sizes. I observed that the download took 0.84 seconds using 8 threads and 25 MB chunks (averaged across 5 test runs). For 8 threads and 100 MB chunks, the download took 1.25 seconds. For 1 thread and 25 MB chunks, the download took 3.23 seconds. Clearly, these parameters greatly influence file transfer times. 

The `io_chunksize` parameter can be set to the maximum possible integer value (`sys.maxsize`), because the vectors are loaded directly to memory without any Disk I/O. All other parameters can be left to their `boto3` defaults, as they do not influence performance as significantly as chunk size and thread count. So, the purpose of this benchmarking experiment will focus on finding strong default values for `multipart_chunksize` and `max_concurrency` for vector download/index upload, that may end up differing from the `boto3` defaults. 

### Testing

[...]

PR #29, opened 2025-03-18
Updated s3 transfer config settings to match benchmarks

rchitale7:
### Description
Updating the s3 file transfer config settings to match the values determined from benchmarking experiments done here: https://github.com/opensearch-project/remote-vector-index-builder/issues/23. I also removed some of the transfer config values that are already set by default by boto3. 

### Issues Resolved
Closes https://github.com/opensearch-project/remote-vector-index-builder/issues/23

### DIFF
--- a/remote_vector_index_builder/core/object_store/s3/s3_object_store.py
+++ b/remote_vector_index_builder/core/object_store/s3/s3_object_store.py
@@ -34,7 +34,6 @@ def get_cpus(factor: float) -> int:
     Returns:
         int: The number of CPUs that will be used for s3 upload or download
 
-    Note: The factor may differ for uploads and downloads, based on benchmarking data
     The cpu count is rounded down to the nearest integer
     """
 
@@ -42,7 +41,7 @@ def get_cpus(factor: float) -> int:
     # if it is none, then default to 1 thread
     cpu_count = os.cpu_count()
     if cpu_count:
-        return math.floor(cpu_count * factor)
+        return max(1, math.floor(cpu_count * factor))
     return 1
 
 
--- a/remote_vector_index_builder/core/object_store/s3/s3_object_store.py
+++ b/remote_vector_index_builder/core/object_store/s3/s3_object_store.py
@@ -9,7 +9,9 @@ import logging
 import os
 import threading
 from functools import cache
+import math
 from io import BytesIO
+import sys
 from typing import Any, Dict
 
 import boto3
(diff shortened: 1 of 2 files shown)


jed326 (approved):

anntians (approved):

remote_vector_index_builder/core/object_store/s3/s3_object_store.py
```diff
@@ -23,6 +25,27 @@
...
+def get_cpus(factor: float) -> int:
+    """Get the number of CPUs to use for s3 upload or download operation
+
+    Args:
+        factor (float): The factor to multiply total cpu count by
+
+    Returns:
+        int: The number of CPUs that will be used for s3 upload or download
+
+    Note: The factor may differ for uploads and downloads, based on benchmarking data
```
owenhalpert, line 37: Is this still considered an open question or has the benchmarking already concluded with the decided values of 0.25 and 0.625?

rchitale7: benchmarking with 16 core CPU showed that 10 threads was best for download, and 4 for upload. Hence the factors of 0.625 and 0.25, respectively.

remote_vector_index_builder/core/object_store/s3/s3_object_store.py
```diff
@@ -23,6 +25,27 @@
...
+    The cpu count is rounded down to the nearest integer
+    """
+
+    # according to mypy, os.cpu_count can be None
+    # if it is none, then default to 1 thread
+    cpu_count = os.cpu_count()
+    if cpu_count:
+        return math.floor(cpu_count * factor)
```
owenhalpert, line 45: What's the minimum CPU count we are assuming here when `os.cpu_count` is not None? Because `cpu_count` < 4 will return 0 for the factor of `0.25`.

remote_vector_index_builder/core/object_store/s3/s3_object_store.py
owenhalpert: Right, I was pointing out that the comment implies there is still benchmarking to be done

remote_vector_index_builder/core/object_store/s3/s3_object_store.py
rchitale7: good catch, let me fix. We'll just default to one thread

owenhalpert: You could do `max(1, math.floor(cpu_count * factor))`

remote_vector_index_builder/core/object_store/s3/s3_object_store.py
rchitale7: will remove to clear the confusion

rchitale7:
Thanks @owenhalpert for catching the bug - i updated the code. 
@jed326 can you take another look? Looks like your old review got dismissed, as expected :)

owenhalpert (approved):

jed326 (approved):

Issue #20, opened 2025-03-04
[Implementation] Create Docker image for Remote Vector Index Builder Core library


rchitale7:
### Description

This issue tracks the creation of the Remote Vector Index Builder `core` Docker image, that will be published on the opensearch-staging dockerhub. This image will contain the core tasks needed to construct an index on a  hardware accelerator (such as GPUs) fleet, for an Opensearch cluster. The tasks are:

- `create_vectors_dataset` - Downloads vector and doc id blobs from a Remote Store repository. The vector and doc id blobs have been previously uploaded by the OpenSearch vector engine. 
- `build_index` - Creates an index on a hardware accelerator fleet, from the vector and doc id blobs
- `upload_index` -  Uploads the index to the Remote Store repository. The Opensearch vector engine will then download this index, and store it on disk for subsequent search requests. 

The `core` Docker image does not offer out of the box APIs that directly integrate with the Opensearch vector engine. Instead, the user is responsible for implementing the data and control plane that provide these APIs, and integrate with `core` under the hood. The vector engine sends build requests to the control plane, which then schedules the requests on data plane `workers`. For a given build request, the worker gets the request parameters, and calls the `core` tasks in sequence. One possible implementation for the worker is given here: https://github.com/opensearch-project/k-NN/issues/2545; however, users are free to choose a different implementation. 

For now, the `core` image will only support building `faiss` indices on GPUs, with the `s3` Remote Store. Future enhancements will be tracked in separate issues. 

The GitHub CI that publishes the `core` image to opensearch staging docker hub will use the GPU 2xlarge runners, to reuse the cached layers of the `faiss-base` image and speed up the builds. However, if are later capacity constrained by integration tests that use the runners, we can utilize the ubuntu-latest default runners for building the `core` image.

rchitale7:
Feature branch: https://github.com/rchitale7/remote-vector-index-builder/tree/api currently has `core` image in `remote_vector_index_builder` folder and `api` image in `testing_api` folder.

Issue #26, opened 2025-03-14
[BUG] CAGRA fails to build index for single document workloads


rchitale7:
### What is the bug?
During testing, we discovered that CAGRA fails to build the index on GPU machines if there's only a single document. Here is the stack trace: 
```
2025-03-13 22:18:11 - core.index_builder.create_gpu_index - INFO - Creating GPU Index.. with IVF_PQ
[W] [22:18:11.804523] Intermediate graph degree cannot be larger than dataset size, reducing it to 1
[W] [22:18:11.804556] Graph degree (32) cannot be larger than intermediate graph degree (0), reducing graph_degree.
using ivf_pq::index_params nrows 1, dim 3, n_lits 1, pq_dim 0
2025-03-13 22:18:12 - core.tasks - ERROR - Error running tasks: C++ exception CUDA error encountered at: file=/opt/conda/include/raft/core/interruptible.hpp line=303: 
Faiss assertion 'err__ == cudaSuccess' failed in virtual faiss::gpu::StandardGpuResourcesImpl::~StandardGpuResourcesImpl() at /tmp/faiss/faiss/gpu/StandardGpuResources.cpp:141; details: CUDA error 700 an illegal memory access was encountered
```
We've observed that this error only occurs when `nrows` (number of documents) = 1. 

As a short term work around, we can configure the remote index build service to reject requests if document count is <= 1. However, this looks like a bug with the faiss CAGRA algorithm that needs to be addressed. 

### How can one reproduce the bug?
1. Upload a float vector and doc id dataset with 1 vector and 1 doc id to an s3 bucket. 
2. Follow the steps here: https://github.com/rchitale7/remote-vector-index-builder/blob/api/DEVELOPER_GUIDE.md to start the remote vector index build service, and trigger a build request.

### What is the expected behavior?
CAGRA should successfully create index for single document workloads. 

### What is your host/environment?
g5.4xlarge EC2 instance, with `Deep Learning Base OSS Nvidia Driver GPU AMI (Amazon Linux 2023)` AMI

### Do you have any screenshots?
_If applicable, add screenshots to help explain your problem._

### Do you have any additional context?
We observed this error while doing integration testing. If this error persists, we will need to update the integration tests to not send single document requests.

anntians:
To reproduce the error, first setup KNN via Developer Guide and checkout main of https://github.com/anntians/k-NN

Then follow the steps below to start remote index builder
```
// 1. Pull GPU docker image
[...]

PR #41, opened 2025-03-25
Added run_tasks function and Dockerfile

rchitale7:
### Description
Created top level `run_tasks` function and Dockerfile for Remote Vector Index Builder core image. `run_tasks` will be the entry point for running the remote index build workflow. Also, consolidated some of the `conftest.py` testing files, and added a validation check to `index_build_parameters` for doc_count > 1 per https://github.com/opensearch-project/remote-vector-index-builder/issues/26

I also tested the code manually on an EC2 instance. 

### Issues Resolved
Resolves https://github.com/opensearch-project/remote-vector-index-builder/issues/20


.github/workflows/publish_remote_core_image.yml
```diff
@@ -0,0 +1,63 @@
...
+  id-token: write
+  contents: read
+
+jobs:
+  build-and-publish-core-image:
+    name: Build and Publish Remote-Vector-Index-Builder Core Image
+    if: github.repository == 'opensearch-project/remote-vector-index-builder'
+    runs-on:
+      group: selfhosted-gpu-runners
+      labels: g62xlarge
```
jed326, line 21: Do we need the GPU runners for this one?

remote_vector_index_builder/core/tasks.py
```diff
@@ -9,24 +9,171 @@
...
+
+            os.remove(index_local_path)
+
+            logger.info(
+                f"Ending task execution for vector path: {index_build_params.vector_path}"
+            )
+            return TaskResult(file_name=os.path.basename(remote_path))
+        except Exception as e:
+            logger.error(f"Error running tasks: {e}")
+            return TaskResult(error=str(e))
```
jed326, line 139: Do we also need to call `vectors_dataset.free_vectors_space()` here?

rchitale7: Yes we do. ill update `free_vectors_space` to be idempotent, and call it in the finally block

.github/workflows/publish_remote_core_image.yml
rchitale7: No we don't, i missed this earlier. Will test with `ubuntu-latest` on feature branch

rchitale7: I'll need to remove this cleanup step as well: https://github.com/opensearch-project/remote-vector-index-builder/pull/41/files#diff-6e7c887dec46aac8dde35e6b6a20bee9e2cd6906e9ab0e2c8ccf73194635d98aR58

rchitale7: updated the issue to mention this: https://github.com/opensearch-project/remote-vector-index-builder/issues/20

jed326: Apologies wanted to discuss on this part a little more:
> Whereas the image layers would be cached in the self hosted group of GPU runners. What do you think about reusing the GPU runners to speed up the builds?

1. Is this caching behavior correct? Don't we want to guarantee we are pulling the latest version of the parent image each time?
2. Aren't we doing the cleanup when building the faiss image anyways: https://github.com/opensearch-project/remote-vector-index-builder/blob/1ad7201948cbb45f8a4bc24fbfdb0c592be036dd/.github/workflows/publish_faiss_base_image.yml#L58-L63
3. How much are we actually speeding up the builds by re-using cached artifacts? We're really only saving on download time of the faiss image right?

rchitale7: So, this line https://github.com/opensearch-project/remote-vector-index-builder/pull/41/files#diff-3df1222cbec3d239d3fc8fb8ca77836152df8ceb6ff389248df8207d1276d4c6R1 is what pulls the faiss base image from opensearch-staging, during the docker build of the core image. If one of the layers of the base image changes, then only that layer needs to be downloaded from opensearch-staging - assuming the other layers are cached. But if the other layers are not cached, then the entire image needs to be re-downloaded. The full image download can take minutes, whereas downloading only the modified layers takes seconds. 

That being said, anytime we make a change to the faiss image, we clean up all the images and cached layers as part of the `publish_faiss_base_image` action. This would result in the core-image action doing a full download, when the core image CI is triggered. I'm not sure if this behavior was intentional or not - @Rajrahane can you comment? I think we can save on minutes on our build times if we remove this clean up step or change it to `docker system prune`. 

cc @navneet1v

jed326: > If one of the layers of the base image changes, then only that layer needs to be downloaded from opensearch-staging - assuming the other layers are cached. But if the other layers are not cached, then the entire image needs to be re-downloaded. The full image download can take minutes, whereas downloading only the modified layers takes seconds.

I'm just trying to make sure the docker cache invalidation actually works the way we want it to. For example, let's say you have 2 machines in your self hosted group. Machine 1 built/published the most recent faiss image, and machine 2 built/published the one before that. Both machines could have the faiss image cached, but does machine 2 always know that it's cache needs to be invalidated in this case? If yes, then this approach looks good to me. (On a lower level, the question would be if cache invalidation is based on TTL or based on checksum?)

> I think we can save on minutes on our build times

These are all post-push actions though, so do we really care about the run time? Especially if we are talking about saving on the order of minutes.

Rajrahane: Docker system prune was suggested by @peterzhuamazon to clean up the g62xlarge GPU Runners 
1. docker system prune is required only for self-hosted gpu runners to keep the gpu-runner clean. If we switch to ubuntu, we can delete the cleanup
2. we can use the g6xlarge rather than the g62xlarge if we want to use the gpu
3. Docker uses layered arch with each layer having a hash, If we use a specific tag, it'll reuse some of the layers and always check remote for changed hashes, and pull those layers

rchitale7: adding on to that - since the core image always checks the remote opensearch staging faiss base image, it will always use the most recent faiss image regardless which machine the build happens on. 

But, I can add back the cleanup step for now, for consistent behavior. And we can revisit this if the build times become a problem. Or, if end up becoming capacity constrained by the integration tests, we can switch this CI to use the ubuntu runners (which cleanup by default, since they are ephemeral). Does that sound ok?

rchitale7: ok, switched to ubuntu runners. I tested on a feature branch and it took 4 mins to build and push to opensearch-staging dockerhub.

jed326 (approved):

.github/workflows/publish_remote_core_image.yml
```diff
@@ -0,0 +1,52 @@
+name: Build and Publish Remote-Vector-Index-Builder Core Image to Docker
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'remote_vector_index_builder/**'
```
Rajrahane, line 8: path should be 'remote_vector_index_builder/core/**'

.github/workflows/publish_remote_core_image.yml
```diff
@@ -0,0 +1,52 @@
...
+      - name: Login to DockerHub
+        uses: docker/login-action@v3
+        with:
+          username: ${{ secrets.REMOTE_VECTOR_DOCKER_USERNAME }}
+          password: ${{ steps.retrieve-values.outputs.dockerhub-password }}
+
+      - name: Push Docker Image
+        run : |
+          docker push opensearchstaging/remote-vector-index-builder:core-1.0.0
+          docker push opensearchstaging/remote-vector-index-builder:core-latest
```
Rajrahane, line 52: Just add one more step for completeness

- name: Runner Cleanups
        if: always()
        run: |
          docker logout

rchitale7: do we really need that though? Since we're using ephemeral ubuntu GitHub runners for this task

jed326 (approved):

Rajrahane (approved):

Issue #42, opened 2025-03-28
[Implementation] Create API handlers for Remote Vector Index Builder Core


rchitale7:
The Remote Vector Index Builder API handler image provides the out the box functionality to use the core library functionality. The API handler follows the specification laid out here: https://github.com/opensearch-project/k-NN/issues/2545 ,and will be used for integration tests on the k-NN repo. This issue tracks the initial creation of the API handler image.

PR #43, opened 2025-03-28
Add API handlers and API Docker image

rchitale7:
### Description
Initial implementation of Remote Vector Index Builder API handlers. The API handler is a Docker image, built from the base Core image. It contains the out the box functionality to integrate Faiss GPU accelerated index building with the k-NN vector engine. 

### Issues Resolved
Closes https://github.com/opensearch-project/remote-vector-index-builder/issues/42


remote_vector_index_builder/app/storage/memory.py
```diff
@@ -0,0 +1,159 @@
...
+        """
+        with self._lock:
+            current_time = datetime.now(timezone.utc)
+            expiration_threshold = current_time - timedelta(seconds=self._ttl_seconds)
+
+            self._store = {
+                job_id: data
+                for job_id, data in self._store.items()
+                if data[1] > expiration_threshold
+            }
```
jed326, line 150: Could you remind me why we are using the TTL like this again? It looks to me like `WorkflowExecutor` still finishes executing the task regardless of if it's in the request store or not so this seems like it would only fail completed jobs?

remote_vector_index_builder/app/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+COPY ./remote_vector_index_builder/app/requirements.txt /remote_vector_index_builder/app/requirements.txt
+
+RUN pip install --no-cache-dir --upgrade -r /remote_vector_index_builder/app/requirements.txt
+
+COPY ./remote_vector_index_builder/app /remote_vector_index_builder/app
+
+ENV PYTHONPATH='${PYTHONPATH}:/tmp/faiss/build/faiss/python:/remote_vector_index_builder'
+RUN ["python", "app/test_imports.py"]
+CMD ["fastapi", "run", "app/main.py", "--port", "80"]
```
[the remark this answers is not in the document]
rchitale7: sure we can make this an environment variable and default to some port > 1024

remote_vector_index_builder/app/storage/memory.py
jed326: So removing the `COMPLETED` and `FAILED` job based on TTL makes sense to me, but the `IN_PROGRESS` one is what I'm stuck on. It looks like `WorkflowExecutor` does not quit out of a given `IN_PROGRESS` task if it is removed from the request store (please correct me if I'm wrong on this), so in that case what does removing `IN_PROGRESS` tasks accomplish? Isn't it better to just let the request store fill up with `IN_PROGRESS` tasks instead of evicting currently running ones, because evicting the `IN_PROGRESS` entries does not free up resources.

rchitale7: Oh yeah i can add a check for that. You're right, removing IN_PROGRESS tasks doesn't really help us with the storage problem.

remote_vector_index_builder/app/services/job_service.py
```diff
@@ -0,0 +1,236 @@
...
+
+        workflow = self._create_workflow(
+            job_id, gpu_mem, cpu_mem, index_build_parameters
+        )
+        logger.info(
+            f"Worker resource status for job id {job_id}: - "
+            f"GPU: {self.resource_manager.get_available_gpu_memory():,} bytes, "
+            f"CPU: {self.resource_manager.get_available_cpu_memory():,} bytes"
+        )
+        self.workflow_executor.submit_workflow(workflow)
```
Rajrahane, line 221: Is this a transaction with none or all actions completed?
What if there is a context switch after line 210 and before 213, and memory is not available in _create_workflow but was available during 208-210?

One possible soln could be 

with allocate_memory: (if this is true)
1. add_to_request_store
2. _create_workflow
3. submit_workflow

remote_vector_index_builder/app/storage/memory.py
jed326: A few more thoughts on this:
* Depending on how retries are implemented client side, you might get into a bad thrashing situation here if we do evict the `IN_PROGRESS` tasks
* IIRC we do want some notion of timeout still though, so maybe this worker could move tasks from `IN_PROGRESS` to `FAILED` or something like that
* Do we need an LRU replacement policy too? For example let's say the request store capacity is 5, and TTL is 10 minutes. If 5 tasks come in and get completed in 1 minute, the request store would still show as full until the 10 minutes are up.

remote_vector_index_builder/app/storage/memory.py
```diff
@@ -0,0 +1,159 @@
...
+    """
+    In-memory implementation of the RequestStore interface.
+
+    This class provides a thread-safe, in-memory storage solution for job requests
+    with configurable maximum size and TTL-based cleanup capabilities.
+
+    Attributes:
+        _store (Dict[str, tuple[Job, datetime]]): Internal dictionary storing jobs and their timestamps
+        _lock (threading.Lock): Thread lock for synchronization
+        _max_size (int): Maximum number of jobs that can be stored
```
Rajrahane, line 32: Can you mention which jobs are considered under the max_size?

If it maintains only the in-progress,
If I have a failed job, or a completed one, and the cleanup process doesn't execute in time, 

If the max jobs is 5, and I have all of them either in completed or failed state
the condition, if len(self._store) >= self._max_size: might return false, where infact there is enough capacity for the InMemoryRequestStore to accept

Maybe using another dictionary for the status API to respond status of job id would help solve this
I can see why keeping the old jobs is needed
I see that the logic uses the same _store dictionary to return status of the job

Possible Solution:
You could set the ttl for upto what time I can find the status of a job and back it up in a processed_store and also
maintaining an unprocessed_store and the max_size can be associated with this unprocessed_store
** I'm Trying to find the TTL on status API if it is mentioned somewhere in the RFC **

Rajrahane:
One comment with memory management.
I understand memory allocations are rounded up to nearest boundary, of say 256 bytes
using floats for multiplications would work, but we'll have to round up to the nearest boundary, and subtract that value from the total_memory used.

We might end up subtracting less and an incorrect request acceptance by the request queue would throw an out of memory error during the index build process

remote_vector_index_builder/app/base/resources.py
```diff
@@ -0,0 +1,107 @@
...
+        Attempt to allocate the specified amount of GPU and CPU memory.
+
+        Args:
+            gpu_memory (float): Amount of GPU memory to allocate, in bytes
+            cpu_memory (float): Amount of CPU memory to allocate, in bytes
+
+        Returns:
+            bool: True if allocation was successful, False if insufficient resources
+        """
+        if not self.can_allocate(gpu_memory, cpu_memory):
```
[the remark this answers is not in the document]
rchitale7: yes this is - i will fix it

remote_vector_index_builder/app/storage/memory.py
jed326: I'm definitely sympathetic to not increasing the scope of this PR, however I do want to make sure we have a clear plan for what we're having in this PR and what will be a fast follow. Without a replacement strategy (LRU or anything else), the configuration described in https://github.com/opensearch-project/k-NN/issues/2545 of `TTL = None` and `capacity = 1` doesn't work because the entry will never get removed.

remote_vector_index_builder/app/services/job_service.py
rchitale7: Yeah let me refactor this a bit to make it more clear. The `_create_workflow` function already tries to allocate the memory, and if it fails, then it throws an error. So technically this is fine but the code is still unclear

remote_vector_index_builder/app/storage/memory.py
rchitale7: So in that situation, I was thinking the control plane that uses this image would be responsible for killing the docker container. Its called out in the "How Request Store feature is intended to be used by user" section of the RFC. TTL and capacity are configurable settings. The default configuration provided in this PR should be suitable for an integration testing harness, but even then, we can improve the configuration by adding stuff like an LRU replacement policy. I just don't think it should be part of this PR.

rchitale7:
The memory estimations are already pretty rough - the formula we're using in `calculate_memory_requirements` isn't actually accurate. Improving the accuracy of memory calculations, along with more robust request store management features, should be taken as follow up issues/PRs, in my opinion.

remote_vector_index_builder/app/storage/memory.py
jed326: Yeah that's fine with me. I think the plan would be:
1. In this PR, don't evict `IN_PROGRESS` tasks
2. Tracking issue for timeout
3. Tracking issue for replacement policy

jed326: Actually I'm not even 100% sure about what the right thing to do for [1] is. Even if we don't evict then it will still get removed as soon as it completes and that could be before the client gets a chance to check the status.

remote_vector_index_builder/app/Dockerfile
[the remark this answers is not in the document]
rchitale7: Had some trouble making this an environment variable. It looks like you can only do that when you pass a string to `CMD` instead of an array. But that would launch fastapi server inside a shell, which i think is undesirable. So i just default to port 1025

remote_vector_index_builder/app/storage/memory.py
rchitale7: per discussion - made change to not evict in progress tasks. For now, we will stick with this replacement policy. But if it causes problems when running ITs, or there is a user request for better caching strategy, we can revisit this.

remote_vector_index_builder/app/services/job_service.py
rchitale7: re-factored the code

rchitale7:
Will add instructions for how to configure and run the API handler once this PR https://github.com/opensearch-project/remote-vector-index-builder/pull/46 is merged. Basically, i will condense whats talked about here https://github.com/opensearch-project/k-NN/issues/2545 into easy to read instructions

remote_vector_index_builder/app/storage/memory.py
```diff
@@ -0,0 +1,159 @@
...
+
+        # Start cleanup thread, TTL seconds is not None
+        if self._ttl_seconds is not None:
+            logger.info(
+                f"Starting cleanup thread for request store with TTL {self._ttl_seconds} seconds"
+            )
+            self._cleanup_thread = threading.Thread(
+                target=self._cleanup_loop, daemon=True
+            )
+            self._cleanup_thread.start()
```
jed326, line 57: I think this is fine for now, but in the future we may want to revisit if it actually makes sense to couple the request store implementation and the replacement strategy. I think ideally a user would be able to plug in LRU vs LFU vs anything else irrespective of whether it's an in memory store or a database or anything else. Just something to think about for "future work"

jed326 (approved):
I will make a second pass on the tests later, but the source code LGTM

remote_vector_index_builder/app/storage/memory.py
rchitale7: Yeah, agreed. We could have `cache` interface (with `lru_cache` as one potential subclass) that `memory` or `database` can then use.

Rajrahane (approved):
LGTM

PR #44, opened 2025-03-28
Fix Memory cleanup issues for GPU and CPU index

Rajrahane:
### Description
Fixes GPU and CPU memory leak issues in Faiss Index Build Service build_index and child methods, build_gpu_index, convert_gpu_to_cpu_index, and write_cpu_index

Some background:
The Index Build Service works in 3 phases
Build a GPU Index
Build a CPU index and delete the GPU Index in memory
Write CPU index to file and delete CPU Index in memory
An IndexIdMap object wraps the vector ids list (c++ vector<> size<1MB for 1mil docs)and the GPU/CPU index (c++ Pointers dynamically allocated) together which is passed to the faiss.write_index method to persist to disc.
Both the GPU and CPU index objects created during the Index Build Process are cleaned up manually and eagerly to free up resources
Bugged code was unable to delete the GPU Index due to memory references lying around somewhere outside python code as per my knowledge
Initially deleting the python object using the python 'del' keyword did not help delete the GPU index
CPU index was being deleted though
Solution
Did some deep dive:
The swigfaiss python binding contains a __swig_destroy__(self) method for every faiss class that calls the c++ destructor
I used them to delete both the GPU Index and CPU Index to standardize the deletion code across the repo
Also explicitly deleted the IndexIdMap for completeness
And made sure and verified in runtime, exception handling code also cleans up GPU/CPU memory
during failures at different stages of the build_index orchestrator method

Screenshot 2025-04-02 at 10 00 44 AM

Screenshot 2025-04-04 at 11 18 11 AM

### Issues Resolved


Rajrahane:
To anyone reviewing this

Was able to cleanup the GPU and CPU Indices taking up memory ~size of dataset being processed. (Dataset tested: 3GB, CPU Index taking RAM 3GB, GPU Index taking GPU memory: ~2GB)
Even after this I notice
Net Change in CPU Memory: ~460MB, GPU Memory: ~264MB
But after the docker process closes, Net change becomes 0, New docker run execution starts with same memory(GPU and CPU) before the first process started. Looks like OS cleaned up afterwards

Rajrahane:
**How to track memory usage?**

**1.CPU:** 

Add “RUN conda install -c conda-forge memory_profiler -y" to remote_vector_index_builder/core/Dockerfile 

The the file that contains the function that needs to be proflled, add the import and a @profile annotation on the function

from memory_profiler import profile

@profile
def my_func():
    a = [1] * (10 ** 6)
    b = [2] * (2 * 10 ** 7)
    del b
    return a

Rebuild the core-latest image and run it

**2.GPU**

In core/requirements.txt add py3nvml on a newline

initialize in main:
nvidia_smi.nvmlInit()

Import in the required file and call the get_gpu_memory()
 from py3nvml import nvidia_smi

def get_gpu_memory():
    handle = nvidia_smi.nvmlDeviceGetHandleByIndex(0)  # GPU 0
    info = nvidia_smi.nvmlDeviceGetMemoryInfo(handle)
    print(f"Total: {info.total/1024**2:.2f}MB")
    print(f"Free: {info.free/1024**2:.2f}MB")
    print(f"Used: {info.used/1024**2:.2f}MB")

Rajrahane:
Low level details and some numbers

I understand the GPU Index size is smaller than the dataset because the GPUIndexConfig is passed the dimensions and sqrt(doc_count) of the vector dataset

Hence when the faiss.GpuIndexCagra object is created it assigned 2GB GPU memory and ~360MB CPU memory

then the IndexIdMap.add_with_ids method trains the gpu index with data
which added ~100MB GPU memory and ~300 CPU/RAM memory

deletion of GPU Index only freed 256MB CPU and 1.6GB GPU memory
which lead to leak of
~264MB GPU memory and
400MB CPU/RAM memory somewhere

Also CPU cleanup leaked 28MB CPU Memory

Even the create_vectors_dataset method managed to leak 28MB during cleanup

when the docker run quit, everything was recovered

Another docker run started the system with the same GPU memory: 429.5 MB and 303MB CPU RAM as the first run

Rajrahane:
Feature issue to include GPU memory/ RAM profiling
https://github.com/opensearch-project/remote-vector-index-builder/issues/45

remote_vector_index_builder/core/common/models/index_builder/faiss/faiss_index_hnsw_cagra_builder.py
```diff
@@ -78,17 +78,16 @@ def convert_gpu_to_cpu_index(
...
+            faiss_gpu_build_index_output.index_id_map.index = None
             # Update the ID map index with the CPU index
             index_id_map = faiss_gpu_build_index_output.index_id_map
             faiss_gpu_build_index_output.index_id_map = None
             index_id_map.index = cpu_index
 
             # Free memory taken by GPU Index
-            del gpu_index
-            del faiss_gpu_build_index_output
+            faiss_gpu_build_index_output.cleanup()
```
jed326, line 90: Does this need to happen in a finally block?

remote_vector_index_builder/core/common/models/index_builder/faiss/faiss_index_hnsw_cagra_builder.py
```diff
@@ -121,7 +120,7 @@ def write_cpu_index(
                 cpu_build_index_output.index_id_map, cpu_index_output_file_path
             )
             # Free memory taken by CPU Index
-            del cpu_build_index_output
+            cpu_build_index_output.cleanup()
```
jed326, line 123: Same here, do we need to do this in finally?

remote_vector_index_builder/core/common/models/index_builder/faiss/faiss_index_hnsw_cagra_builder.py
```diff
@@ -78,17 +78,16 @@ def convert_gpu_to_cpu_index(
             cpu_index.base_level_only = self.base_level_only
 
             # Copy GPU index to CPU index
-            gpu_index = faiss_gpu_build_index_output.gpu_index
-            gpu_index.copyTo(cpu_index)
+            faiss_gpu_build_index_output.gpu_index.copyTo(cpu_index)
 
+            faiss_gpu_build_index_output.index_id_map.index = None
```
rchitale7, line 83: do you need lines 83 and 86? It looks like we already do these steps int he `cleanup` function

Rajrahane: Yes
The cleanup method must only delete the GPU Index in the good case

83-86 will remove reference of the IndexIdMap from the gpu_build_index_output.
83 to 86 are needed to prevent the second part of GPU cleanup from running, or else it will delete the IndexIdMap currently being used for CPU

remote_vector_index_builder/core/common/models/index_builder/faiss/faiss_index_hnsw_cagra_builder.py
Rajrahane: No
To me it feels like its part of the CPU Index process
The finally block would run irrespective of whether the CPU Index was created and delete the GPU Index, which is not the expected outcome.
Finally should be used for resource closure. This is more of a output of some other method
We could setup a write GPU to disc or some sort of fallback instead of cleaning it from GPU if the CPU Index fails
Hence I feel it needs to be in the try section where it'll cleanup only when CPU index creation succeeds.

And if the process fails, the top level build_index in faiss_index_build_service would call the cleanup in the Exception handler to clean up (https://github.com/opensearch-project/remote-vector-index-builder/blob/dcc5f9e92fc6cd44bcb6f1a8a14dd5c8cc643925/remote_vector_index_builder/core/index_builder/faiss/faiss_index_build_service.py#L106-L112).

remote_vector_index_builder/core/common/models/index_builder/faiss/faiss_index_hnsw_cagra_builder.py
rchitale7: Imo it would be more clear if you separate this cleanup function into clean_up_gpu_index and clean_up_id_map functions. And then call them independently. Is it possible to do that?

Rajrahane:
Added a few more comments to the cleanup methods and the GPU to CPU convertor method for code flow clarity

rchitale7 (approved):

navneet1v (approved):

Issue #31, opened 2025-03-21
[BUG] FilteredSearchANNSearchIT Failure When Building Index Remotely


anntians:
### What is the bug?
Very similar bug to https://github.com/opensearch-project/remote-vector-index-builder/issues/30. 

When testing the KNN plugin against remote index builder, the test `testFilteredSearchWithFaissHnsw_whenFiltersMatchAllDocs_thenReturnCorrectResults` failed due to a read time out after sending a build request to the remote index builder.

Looking at the logs for remote index builder, it rejects (422) several build requests coming from the `addKnnDocWithAttributes` method when adding test data, which is expected as remote builder will reject build requests with `doc_count <= 1` (ref: https://github.com/opensearch-project/remote-vector-index-builder/issues/26). Then it accepts the build request coming from the `forceMergeKnnIndex` method, and then is stuck, causing the test to time out and fail.

Previously the index builder freezes due to incorrect test data, such as wrong dimensions/data type. However, the `.knndid` and `.knnvec` file size look correct with `.knnvec` 3x the size of `.knndid` due to the test vector dimensions are 3.

Remote index builder logs:
```
2025-03-18 01:19:27 - testing_api.app.main - INFO - Error while validating parameters: #[{'field': 'body.doc_count', 'message': 'Input should be greater than 1', 'type': 'greater_than'}]
      INFO   172.31.37.21:34022 - "POST /_build HTTP/1.1" 422
2025-03-18 01:19:27 - testing_api.app.main - INFO - Error while validating parameters: #[{'field': 'body.doc_count', 'message': 'Input should be greater than 1', 'type': 'greater_than'}]
      INFO   172.31.37.21:34022 - "POST /_build HTTP/1.1" 422
2025-03-18 01:19:27 - testing_api.app.main - INFO - Error while validating parameters: #[{'field': 'body.doc_count', 'message': 'Input should be greater than 1', 'type': 'greater_than'}]
      INFO   172.31.37.21:34022 - "POST /_build HTTP/1.1" 422
2025-03-18 01:19:27 - testing_api.app.main - INFO - Error while validating parameters: #[{'field': 'body.doc_count', 'message': 'Input should be greater than 1', 'type': 'greater_than'}]
      INFO   172.31.37.21:34022 - "POST /_build HTTP/1.1" 422
[...]

Issue #30, opened 2025-03-21
[BUG] KNNMapperSearcherIT Failure When Building Index Remotely


anntians:
### What is the bug?
While testing the KNN plugin `KNNMapperSearcherIT` against the remote index builder, the tests `testKNNResultsUpdateDocAndForceMerge` and `testKNNResultsWithForceMerge` failed due to a read time out after sending a build request to the remote index builder. 

Looking at the logs for remote index builder, it rejects (422) several build requests coming from the `addKnnDoc` method when adding test data, which is expected as remote builder will reject build requests with `doc_count <= 1` (ref: https://github.com/opensearch-project/remote-vector-index-builder/issues/26). Then it accepts the build request coming from the `forceMergeKnnIndex` method, and then is stuck, causing the tests to time out and fail.

Previously the index builder freezes due to incorrect test data, such as wrong dimensions/data type. However, the `.knndid` and `.knnvec` file size look correct with `.knnvec` 2x the size of `.knndid` due to the test vector dimensions are 2.

Remote index builder logs:
```
2025-03-17 22:47:07 - testing_api.app.main - INFO - Error while validating parameters: #[{'field': 'body.doc_count', 'message': 'Input should be greater than 1', 'type': 'greater_than'}]
      INFO   172.31.37.21:41794 - "POST /_build HTTP/1.1" 422
2025-03-17 22:47:07 - testing_api.app.main - INFO - Error while validating parameters: #[{'field': 'body.doc_count', 'message': 'Input should be greater than 1', 'type': 'greater_than'}]
      INFO   172.31.37.21:41794 - "POST /_build HTTP/1.1" 422
2025-03-17 22:47:07 - testing_api.app.main - INFO - Error while validating parameters: #[{'field': 'body.doc_count', 'message': 'Input should be greater than 1', 'type': 'greater_than'}]
      INFO   172.31.37.21:41794 - "POST /_build HTTP/1.1" 422
2025-03-17 22:47:07 - testing_api.app.main - INFO - Error while validating parameters: #[{'field': 'body.doc_count', 'message': 'Input should be greater than 1', 'type': 'greater_than'}]
      INFO   172.31.37.21:41794 - "POST /_build HTTP/1.1" 422
[...]

PR #52, opened 2025-04-12
Increase the minimum doc count to 5

rchitale7:
### Description
During integration testing, we observed that CAGRA would fail to build an index when the doc count was 1, 3 or 4. Specifically, the `copyTo` method used to convert the GPU index to CPU index would hang. This would cause certain kNN integration tests to fail - see https://github.com/opensearch-project/remote-vector-index-builder/issues/31 and https://github.com/opensearch-project/remote-vector-index-builder/issues/30. We tested up to having 23,000 docs, and did not see this issue with any other doc count. So, I believe it is sufficient to raise the minimum doc count threshold to 5, to prevent this issue from happening. 

I also refactored a lot of the unit tests to make testing these types of changes easier in the future

### Issues Resolved
Resolves https://github.com/opensearch-project/remote-vector-index-builder/issues/31 and https://github.com/opensearch-project/remote-vector-index-builder/issues/30. 


remote_vector_index_builder/core/common/models/index_build_parameters.py
```diff
@@ -144,7 +144,7 @@ class IndexBuildParameters(BaseModel):
     doc_id_path: str
     tenant_id: str = ""
     dimension: int = Field(gt=0)
-    doc_count: int = Field(gt=1)
+    doc_count: int = Field(gt=4)
```
navneet1v, line 147: please update the API contracts too

rchitale7: updated the docs

navneet1v:
@rchitale7 is the branch properly rebased. The number of changes this small change has introduced seems to be pretty high.

API.md
```diff
@@ -41,6 +41,7 @@ Request Response:
 * The top two parameters are derived by the KNN plugin from the customer's repo setting. `container_name` is used to specifically refer to the "bucket" (or non-S3 equivalent), rather than the name of the repository itself.
 * Tenant ID is included to be used for billing, authorization, etc
 * Dimension, doc count, and data type are specifically placed on the first JSON level so the build service can quickly use them first to calculate workload size
+* Doc count must be greater than 4, since doc counts of 1, 3, and 4 have been observed to consistently fail during index builds
```
jed326, line 44: Not blocking this PR, but can we link the FAISS issue somewhere? https://github.com/facebookresearch/faiss/issues/4260

navneet1v (approved):

jed326 (approved):

Issue #49, opened 2025-04-11
[Discussion] GPU memory usage spikes while building FAISS index


rchitale7:
### Background

While performing the GPU benchmarking experiments called out here: https://github.com/opensearch-project/remote-vector-index-builder/issues/36, I noticed something strange: the peak GPU memory usage during the CAGRA Index Build would spike to about double the size of the testing dataset. For example, for this 6.144 GB dataset: https://huggingface.co/datasets/navneet1v/datasets/resolve/main/open-ai-1536-temp.hdf5?download=true, the GPU memory spiked to as high as 10.5 GB, when I monitored the GPU usage with `nvidia-smi` in the background. Even after reducing the `trainset_ratio` from 0.5 -> 0.1 and `refine_rate` from 2 -> 1 (these are CAGRA hyper parameters that impact GPU memory usage - if they are lower, less memory is consumed), the peak memory was 9.3 GB. In the benchmarking code, the full dataset is first loaded into CPU memory as a `numpy` array, before it is used in the CAGRA index build. 

However, after I modified the code to store the dataset on disk, and create a memory-mapped array using `numpy.mmap`, I saw the peak GPU memory drop. For the same 6.144 GB dataset, with `trainset_ratio` set to 0.5 and `refine_rate` set to 2 and using a memory-mapped dataset array, the peak GPU memory was 5.039. After lowering `trainset_ratio` to 0.1 and `refine_rate` to 1, the peak GPU memory was 3.56 GB. So, there is roughly a 5.5 GB drop in peak GPU memory using the memory-mapped approach, compared with loading the dataset fully into CPU memory. That is about the same size as the dataset. So, what is likely happening is that if the dataset gets fully loaded into CPU memory, it gets copied to GPU memory. However, this is not expected to occur, because the CAGRA `attach_dataset_on_build` parameter is set to `False`. This parameter is supposed to prevent loading the dataset into GPU memory: https://docs.rapids.ai/api/cuvs/stable/cpp_api/neighbors_cagra/#_CPPv423attach_dataset_on_build (Note that FAISS calls this parameter `store_dataset`, but it still maps 1:1 to `attach_dataset_on_build`). We have an open issue with FAISS regarding this: https://github.com/facebookresearch/faiss/issues/4274

### Problem

[...]

PR #57, opened 2025-04-18
Add mmap based storage of vectors, while keeping buffered storage as default

rchitale7:
### Description
This issue enables the vector and doc id binary downloaded from Remote Store to be saved to disk, and read using mmap. The caller of `run_tasks` can download the vectors to disk if they set `storage_mode` = "disk", but the default storage mode is still `memory`. I refactored the code so that we can support other storage modes if necessary. Key to the refactoring is the `BinarySource` object that wraps the storage mechanism; `FileSource` wraps the file object used for `numpy.mmap`, while `BufferSource` wraps the buffer object used for `numpy.frombuffer`. 

I verified manually that both storage modes work, and create the expected faiss graph.

### Issues Resolved
Partially resolves for https://github.com/opensearch-project/remote-vector-index-builder/issues/49. I need to still update the `USER_GUIDE.md` to specify that `memory` is the default approach, but with CUDA versions >= 12, we see memory spike. So to avoid memory spike, user can use `disk` (at the expense of slower index builds)

### DIFF
--- a/remote_vector_index_builder/core/binary_source/file_source.py
+++ b/remote_vector_index_builder/core/binary_source/file_source.py
@@ -1,0 +1,40 @@
+# Copyright OpenSearch Contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+
+from core.binary_source.binary_source import BinarySource
+from core.object_store.object_store import ObjectStore
+
+from tempfile import TemporaryDirectory
+import numpy as np
+from numpy.typing import NDArray
+
+
+class FileSource(BinarySource):
+    def __init__(self):
+        self._temp_dir = None
+        self._source = None
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        if self._source is not None:
+            self._source.close()
+        if self._temp_dir is not None:
+            self._temp_dir.cleanup()
+
+    def read_from_object_store(
+        self, object_store: ObjectStore, object_store_path: str
+    ) -> None:
+        filename = object_store_path.replace("/", "_").replace("\\", "_")
+        self._temp_dir = TemporaryDirectory()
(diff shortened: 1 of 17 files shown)


remote_vector_index_builder/core/binary_source/file_source.py
```diff
@@ -0,0 +1,40 @@
...
+            self._source.close()
+        if self._temp_dir is not None:
+            self._temp_dir.cleanup()
+
+    def read_from_object_store(
+        self, object_store: ObjectStore, object_store_path: str
+    ) -> None:
+        filename = object_store_path.replace("/", "_").replace("\\", "_")
+        self._temp_dir = TemporaryDirectory()
+        self._source = open(f"{self._temp_dir.name}/{filename}", "wb+")
```
jed326, line 35: Why `wb+` instead of `wb`? Do we expect to be writing multiple things to this file?

rchitale7: I thought the `+` is to enable reading of the file. We write to the file when downloading the vectors, but CAGRA needs to be able to read the vectors when constructing the graph.

jed326: https://docs.python.org/3/library/functions.html#open

Mostly I was asking if we should truncate the file, or do an assertion on the file being empty. Seems like `wb+` does truncate the file though.

jed326 (dismissed):

rchitale7:
Based on discussion in https://github.com/facebookresearch/faiss/issues/4274, we determined that the reason the memory spikes when storing vectors in CPU memory is because of the CUDA version. This issue is observed with CUDA version >= 12. I've updated this PR to make memory based storage the default, while still giving the caller the option to use disk/mmap based storage as another option. I'll update the `USER_GUIDE.md` to specify when to use disk v.s. memory storage after this PR is merged.

jed326 (approved):

navneet1v:
@rchitale7 fix the commit message and also the PR description to correctly point out what the PR is doing

Issue #19, opened 2025-03-04
[Implementation] Add integration tests to remote-vector-index-builder core


rchitale7:
### Description

This issue tracks implementation of integration tests for the remote-vector-index-builder `core` library. 

- [ ] `create_vectors_dataset`
    - [ ] Validate that vector and doc id blobs are successfully read from s3 into numpy array, for 1 and 5 GB vector blobs
    - [ ] Validate that vector and doc id blobs do not get read, when checksums are invalid
    - [ ] Validate that vector and doc id blobs do not get read, when credentials are invalid
- [ ] `build_gpu_index`
    - [ ] Validate that index is created, for 1 and 5 GB vector blobs
- [ ] `upload_index`
    - [ ] Validate that index is successfully uploaded to s3, for 1 and 5 GB vector blobs
    - [ ] Validate that index does not get uploaded, when checksums are invalid
    - [ ] Validate that index does not get uploaded, when credentials are invalid
- [ ] End-to-end integration test that calls `create_vectors_dataset`, `build_gpu_index`, and `upload_index` in sequence

krisfreedain:
[Catch All Triage](https://github.com/opensearch-project/.github/blob/main/TRIAGING.md) - [1](https://github.com/krisfreedain) [2](https://github.com/andrross)

Rajrahane:
Ref:
Override S3_ENDPOINT https://github.com/opensearch-project/remote-vector-index-builder/pull/47

PR #58, opened 2025-04-18
Add e2e test setup

Rajrahane:
### Description
Executing the E2E core functionality test with dummy datasets
Datasets Generated: 

| Dimensions   | #Vectors      | Size in GB |
|--------------|-------------|-----------|
| 128                 | 1_000_000      | 0.48         |
| 384 | 1_000_000 | 1.43 |
| 768 | 1_000_000 | 2.86 |

AWS S3 functionality can be mocked using [localstack](https://docs.localstack.cloud/getting-started/)

Setting up a localstack and e2e-tests docker container to execute the test

Tasks executed:
- Create Test Bucket, Upload dummy vectors to S3(localstack)
- Run full indexing flow (run_tasks): Download Vectors, Create Index, Upload Index
- S3 Teardown: Delete S3 artifacts, test bucket
- Docker Teardown (Github CI): Prune system: unused volumes, images, network

### Issues Resolved
https://github.com/opensearch-project/remote-vector-index-builder/issues/19


Rajrahane:
Github CI Test Run executed in 12:26 secs

.github/workflows/remote-vector-index-builder/run-tests.yml
rchitale7, line 1: why are we deleting this file

e2e/test_core/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM opensearchstaging/remote-vector-index-builder:core-latest
```
rchitale7, line 1: I think this should still use the core image we build locally, right? Assuming we tag the image before with `opensearchstaging/remote-vector-index-builder:core-latest`

.github/workflows/remote-vector-index-builder/run-tests.yml
Rajrahane: old file in wrong folder
missed by me 
original file is present as well under .github/workflows

e2e/test_core/run_e2e.py
```diff
@@ -0,0 +1,103 @@
...
+
+    # run for each dataset in yml e2e test vector index builder
+    # Teardown dataset and sucess artfacts from memory and s3
+
+    # Ideally run this on a container so that memory gets freed up at end of tests
+    # Assert docker container quit with exit 0 -> all tests suite exitted normally
+    dataset_generator = VectorDatasetGenerator(config_path)
+
+    try:
+        # Creat etest bucket if it doesn't exist
```
jed326, line 28:
```suggestion
        # Create test bucket if it doesn't exist
```

e2e/test_core/run_e2e.py
```diff
@@ -0,0 +1,103 @@
...
+from core.object_store.types import ObjectStoreType
+from e2e.test_core.vector_dataset_generator import VectorDatasetGenerator
+from botocore.exceptions import ClientError
+from core.tasks import run_tasks
+import os
+
+
+def run_e2e_index_builder(config_path: str):
+
+    # run for each dataset in yml e2e test vector index builder
```
jed326, line 20: Just point to the actual file

e2e/test_core/run_e2e.py
```diff
@@ -0,0 +1,103 @@
...
+                    "retries": s3_config["retries"],
+                    "region": s3_config["region"],
+                    "S3_ENDPOINT_URL": os.environ.get(
+                        "S3_ENDPOINT_URL", "http://localhost:4566"
+                    ),
+                }
+                result = run_tasks(
+                    index_build_params=index_build_params,
+                    object_store_config=object_store_config,
+                )
```
jed326, line 74: I was under the impression that we wanted to test the API layer here, did that change? Is our plan to test the API layer separately? cc @rchitale7

rchitale7: I though testing the `run_tasks` function in core was sufficient, since I anticipate there being less changes in the `api` folder. If not, using the python `requests` library to start the build and then check the build in a while loop should work.

jed326: The API layer is in use for the tests run in the k-NN repo as well though. I am fine if you think only unit test coverage for the API layer is sufficient I was just under the impression we were testing via API calls here

e2e/test_core/run_e2e.py
```diff
@@ -0,0 +1,103 @@
...
+    # Assert docker container quit with exit 0 -> all tests suite exitted normally
+    dataset_generator = VectorDatasetGenerator(config_path)
+
+    try:
+        # Creat etest bucket if it doesn't exist
+        s3_client = dataset_generator.object_store.s3_client
+        bucket = dataset_generator.config["storage"]["s3"]["bucket"]
+        try:
+            s3_client.create_bucket(Bucket=bucket)
+            print(f"Created bucket: {bucket}")
```
rchitale7, line 33: nit: use `logging` module instead of `print` statements

peterzhuamazon:
Talked to @Rajrahane offline and in order to not pull base image every time, which take a long time for a 15G image.

We can switch the system prune commands to individual parts:
```
docker container prune -f
docker volume prune -af
docker network prune -f
docker rmi $(docker images -f "dangling=true" -q) || echo clean
```

The last command will only remove dangling images but keeps tagged image untouched.
Since you are always using the `faiss-base-latest` tag, even if the image get updated in remote, the old image on dist will be replaced with the new image on the same tag, effectively making the old image dangling. And will be cleaned up by the last command.

Thanks.

peterzhuamazon:
> > Talked to @Rajrahane offline and in order to not pull base image every time, which take a long time for a 15G image.
> > We can switch the system prune commands to individual parts:
> > ```
> > docker container prune -f
> > docker volume prune -af
> > docker network prune -f
> > docker rmi $(docker images -f "dangling=true" -q) || echo clean
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The last command will only remove dangling images but keeps tagged image untouched.
> > Since you are always using the `faiss-base-latest` tag, even if the image get updated in remote, the old image on disk will be replaced with the new image on the same tag, effectively making the old image dangling. And will be cleaned up by the last command.
> > Thanks.
> 
> @peterzhuamazon what will be the way to update the base image if there are changes in the base faiss image? any way we can force the update on the docker pull?

In docker build step add a param `--pull` to always pull remote even if local has a copy. If updated in remote then it will update local properly, else use local.

.github/workflows/run_e2e_tests.yml
```diff
@@ -0,0 +1,59 @@
...
+      group: selfhosted-gpu-runners
+      labels: g6xlarge
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Build Docker Image
+        run : |
+          docker build  -f ./e2e/test_core/Dockerfile . -t opensearchstaging/remote-vector-index-builder:e2e-latest
```
peterzhuamazon, line 40:
```suggestion
          docker build --pull  -f ./e2e/test_core/Dockerfile . -t opensearchstaging/remote-vector-index-builder:e2e-latest
```

.github/workflows/run_e2e_tests.yml
```diff
@@ -0,0 +1,59 @@
...
+      
+      - name: Run E2E Tests
+        run : |
+            docker compose -f ./e2e/test_core/docker-compose.yml up --exit-code-from remote-vector-index-builder
```
peterzhuamazon, line 44:
```suggestion
            docker compose pull
            docker compose -f ./e2e/test_core/docker-compose.yml up --exit-code-from remote-vector-index-builder
```

.github/workflows/run_e2e_tests.yml
```diff
@@ -0,0 +1,59 @@
...
+
+      - name: Docker-Compose cleanup
+        if: always()
+        run : |
+            docker compose -f ./e2e/test_core/docker-compose.yml -v down
+
+      - name: Runner Cleanups
+        if: always()
+        run : | 
+          docker system prune -af --volumes
```
peterzhuamazon, line 54:
```suggestion
          docker container prune -f
          docker volume prune -af
          docker network prune -f
          docker rmi $(docker images -f "dangling=true" -q) || echo clean
```

.github/workflows/run_e2e_tests.yml
peterzhuamazon: Always pull remote + do not use opensearchstaging if it is a local image.

.github/workflows/run_e2e_tests.yml
peterzhuamazon: Always pull remote for the localstack/localstack.

e2e/test_core/docker-compose.yml
```diff
@@ -0,0 +1,39 @@
...
+    environment:
+      - SERVICES=s3
+      - AWS_ACCESS_KEY_ID=test
+      - AWS_SECRET_ACCESS_KEY=test
+      - AWS_DEFAULT_REGION=us-east-1
+    networks:
+      - vector-network
+  remote-vector-index-builder:
+    container_name: remote-vector-index-builder
+    image: opensearchstaging/remote-vector-index-builder:e2e-latest
```
peterzhuamazon, line 18:
```suggestion
    image: remote-vector-index-builder:e2e-latest
```

peterzhuamazon: local e2e test do not need opensearchstaging

.github/workflows/run_e2e_tests.yml
Rajrahane: done

Rajrahane:
Solution worked!
faiss-base and localstack sha was reused from cache

e2e/test_core/run_e2e.py
Rajrahane: added a default value pointing to the file

e2e/test_core/run_e2e.py
Rajrahane: skip API tests for now @rchitale7 ?

.github/workflows/run_e2e_tests.yml
Rajrahane: skipped this unnecessary line
default pull strategy on docker-compose is always

e2e/test_core/run_e2e.py
rchitale7: lets do this in a follow up PR. Should be able to test the api code in a similar manner to the core code.

.github/workflows/run_e2e_tests.yml
```diff
@@ -51,9 +51,14 @@ jobs:
       - name: Runner Cleanups
         if: always()
         run : | 
-          docker system prune -af --volumes
+          docker container prune -f
+          docker volume prune -af
+          docker network prune -f
+          docker rmi $(docker images -f "dangling=true" -q) || echo clean
```
rchitale7, line 57: I know the faiss base image doesn't get updated very frequently but technically this line here: https://github.com/opensearch-project/remote-vector-index-builder/blob/main/.github/workflows/publish_faiss_base_image.yml#L59 will undo the benefits we get from this change, right? cc @peterzhuamazon

rchitale7 (approved):

e2e/test_core/run_e2e.py
jed326: Can you create a tracking issue for this?

.github/workflows/run_e2e_tests.yml
peterzhuamazon: Not necessarily if you dont push image that often.
It depends if your team wants to change that workflow to also onboard the new cleanup method.
In my original proposal I am just to ensure runners are properly cleanup so that each run is not affected by previous run.

jed326 (approved):

rchitale7 (approved):

PR #68, opened 2025-05-22
Added paths to trigger core and api image builds when base image changes

navneet1v:
### Description
Added paths to trigger core and api image builds when base image changes.

Reason for adding this is, whenever there is a change in base image, the core and api images which are published should also be updated since base is changed. If we don't do this, then published core and api images will be stale.

Similar thing is done for api image too, where api image is rebuild if changes in core or base image is there.

Thanks @rchitale7 for suggesting this.

### Issues Resolved
NA

### DIFF
--- a/.github/workflows/publish_remote_api_image.yml
+++ b/.github/workflows/publish_remote_api_image.yml
@@ -7,6 +7,7 @@ on:
     paths:
       - 'remote_vector_index_builder/app/**'
       - '.github/workflows/publish_remote_api_image.yml'
+
   workflow_call: # enables workflow to be reused
     secrets:
       REMOTE_VECTOR_DOCKER_USERNAME:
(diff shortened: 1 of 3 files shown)


.github/workflows/publish_remote_api_image.yml
```diff
@@ -7,6 +7,11 @@ on:
     paths:
       - 'remote_vector_index_builder/app/**'
       - '.github/workflows/publish_remote_api_image.yml'
+      - 'remote_vector_index_builder/core/**'
```
rchitale7, line 10: Technically we don't need this since we auto publish the api image, anytime theres change in core: https://github.com/opensearch-project/remote-vector-index-builder/blob/5e0f1de7eabd9ac90ff91e7eb43bbdb62bc764ad/.github/workflows/publish_remote_core_image.yml#L60

navneet1v: fixed

rchitale7 (approved):

jed326 (approved):

PR #75, opened 2025-06-05
Add benchmarking tool

rchitale7:
### Description
Adding the GPU/CPU Faiss benchmarking tool to the repository. Instructions for running the tool are in `benchmarking/README.md`


navneet1v (dismissed):

navneet1v (dismissed):

benchmarking/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM opensearchstaging/remote-vector-index-builder:core-latest
```
Rajrahane, line 1: core-snapshot

rchitale7: We eventually want to run this benchmark harness automatically for snapshot and latest images. So i would say we should figure out a way to make the image tag an environment variable. But can come back to this later

Issue #77, opened 2025-06-06
[Discussion] Image Versioning Strategy for Remote-Vector-Index-Builder


Rajrahane:
### Proposed Versioning Strategy:
The following mentions the versioning strategy for images published to opensearchstaging DockerHub and ECR repos

#### A) For Merges to main
Images should be tagged with the -snapshot tag. These include the following images -
1. faiss-base-snapshot
2. core-snapshot
3. api-snapshot
4. e2e-snapshot (not published)
Images are not deleted from the Github CI runner since the frequency of building the child images from the faiss-base and core images is high. This reduces the total image build time.

Regular cleanup of stale docker artifacts on Github CI runner takes place when faiss-base-snapshot image is created.
**Please see the cleanup step in [publish_faiss_base_image.yml](https://github.com/opensearch-project/remote-vector-index-builder/blob/main/.github/workflows/publish_faiss_base_image.yml)**

k-nn repository should run [remote_index_build.yml](https://github.com/opensearch-project/k-NN/blob/main/.github/workflows/remote_index_build.yml) with the api-snapshot image

#### B) Version Branches (eg: 1.0.0)
Images should be tagged with a -latest and -major.minor.patch tag These include the following images
1. faiss-base-latest, faiss-base-1.0.0
2. core-latest, core-1.0.0
3. api-latest, api-1.0.0

Future versions can update/overwrite the -latest image on opensearchstaging
Github CI will be triggered manually using workflow-dispatch
Both types of images will also be deleted from the Github CI runner during cleanup.

**The Images created are later released to opensearchproject during a Version Release**

Task Breakdown

- [x] Update Current Github CIs to publish '-snapshot' images rather than '-latest'
- [x] Add a new Github CI to publish Version Images to opensearchstaging
- [ ] Add a Jenkins Pipeline to push images to opensearchproject

PR #83, opened 2025-06-12
Add Release CI

Rajrahane:
### Description
Create Release CI and Jenkinsfile to release VERSION and Latest images to opensearchproject

### Issues Resolved
Partly completes https://github.com/opensearch-project/remote-vector-index-builder/issues/77


Rajrahane:
@peterzhuamazon It would be great if you could take a look at the jenkinsfile for me

specifically the runner config and post success, updating the release to draft=false

.github/workflows/release.yml
```diff
@@ -0,0 +1,77 @@
...
+      steps:
+        - name: Get Approvers
+          id: get_approvers
+          run: |
+            echo "approvers=$(cat .github/CODEOWNERS | grep @ | tr -d '* ' | sed 's/@/,/g' | sed 's/,//1')" >> $GITHUB_OUTPUT
+
+        - name: Get Manual Workflow Approval
+          uses: trstringer/manual-approval@v1
+          with: 
+            secret: ${{ github.TOKEN }}
```
Gankris96, line 40: just curious - why is `TOKEN` in uppercase? I am assuming you are reading it from here https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context

Then shouldn't it be lower case?

Gankris96, line 40: just curious - why is TOKEN in uppercase? I am assuming you are reading it from here https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context

Then shouldn't it be lower case? Or is it case insensitive?

.github/workflows/release.yml
```diff
@@ -0,0 +1,77 @@
...
+            approvers: ${{ steps.get_approvers.outputs.approvers }}
+            minimum-approvals: 3
+            exclude-workflow-initiator-as-approver: false
+            issue-title: 'Release Remote-Vector-Index-Builder Version v${{ github.event.inputs.version }} Images to opensearchproject '
+            issue-body: >
+                Please approve or deny release of Remote-Vector-Index-Builder 
+                Version v${{ github.event.inputs.version }} Images to opensearchproject
```
Gankris96, line 47: Here it says Version `v${{ github.event.inputs.version }}`

But in below in Create Release Description it says 
```
echo 'version: ${{ github.event.inputs.version }}' > release-description.yaml
```

So will the version be with `v` or without ? 

Not sure if this is a big deal but just wanted to check -

Rajrahane: The file will have it w/o the v
This will help parse the whole version string directly

rchitale7 (approved):
Approving, but would be great if @peterzhuamazon or @rishabh6788 could take a look, especially the jenkins release file.

.github/workflows/release.yml
```diff
@@ -0,0 +1,77 @@
...
+        type: string
+        required: true
+      release_latest_tag:
+            description: 'Publish latest tag'
+            required: true
+            type: boolean
+      release_version_tag:
+            description: 'Publish version tag'
+            required: true
+            type: boolean
```
peterzhuamazon, line 17: Why use workflow dispatch when we should trigger by cutting a tag.
If you do this way there is no prevention of re-releasing multiple on the same commit.
But also cannot trigger the jenkins workflow because it is listening to the release creation event.

peterzhuamazon (commented):
I feel like a lot of these can be fine-tuned especially the referenced data prepper is too old.

Let's have a talk on this.

Thanks.

peterzhuamazon (approved):

PR #98, opened 2025-07-17
Add AWS creds config to object_store

Rajrahane:
### Description
Add AWS Creds Config to Object Store
Implements an S3ClientConfig pydantic class to accept boto3 client args
Ensures model variables data type validation.

### Issues Resolved


remote_vector_index_builder/app/services/index_builder.py
```diff
@@ -36,7 +37,13 @@ def build_index(
         """
         s3_endpoint_url = os.environ.get("S3_ENDPOINT_URL", None)
         result = run_tasks(
-            workflow.index_build_parameters, {"S3_ENDPOINT_URL": s3_endpoint_url}
+            workflow.index_build_parameters,
+            {
+                "s3_client_config": S3ClientConfig(
+                    region_name=os.environ.get("REGION_NAME", "us-west-2"),
```
Rajrahane, line 43: Wondering whether we should set it as AWS_DEFAULT_REGION
Thoughts?

Rajrahane:
Note: https://github.com/anntians/k-NN/blob/main/.github/workflows/remote_index_build.yml
Although the REGION parameter is set to default
we would like to add it as an env here

Also suggest changing the CI to run in docker-compose instead of multiple docker statements
Adding new env's is oneliner and cleaner to review

e2e/api/docker-compose.yml
```diff
@@ -25,6 +25,7 @@ services:
       - AWS_ACCESS_KEY_ID=test
       - AWS_SECRET_ACCESS_KEY=test
       - AWS_DEFAULT_REGION=us-east-1
+      - REGION_NAME=us-east-1
```
rchitale7, line 28: why is this duplicated with `AWS_DEFAULT_REGION`?

remote_vector_index_builder/core/object_store/s3/s3_object_store.py
```diff
@@ -62,12 +61,15 @@ def get_boto3_client(
...
-    config = Config(retries={"max_attempts": retries})
+    config = Config(retries={"max_attempts": s3_client_config.max_retries})
     return boto3.client(
         "s3",
         config=config,
-        region_name=region,
-        endpoint_url=endpoint_url,
+        region_name=s3_client_config.region_name,
+        endpoint_url=s3_client_config.endpoint_url,
+        aws_access_key_id=s3_client_config.aws_access_key_id,
```
rchitale7, line 70: what happens when access key, secret key, session token are null? Our ITs run with local stack, but what happens when we use a real s3 bucket?

remote_vector_index_builder/core/object_store/s3/s3_object_store_config.py
```diff
@@ -0,0 +1,24 @@
...
+    endpoint_url: Any | None = None
+    max_retries: int = 3
+    aws_access_key_id: Optional[str] = None
+    aws_secret_access_key: Optional[str] = None
+    aws_session_token: Optional[str] = None
+
+    def __hash__(self):
+        return hash((self.region_name, self.max_retries, self.endpoint_url))
+
+    def __eq__(self, other):
```
rchitale7, line 17: is there a reason why we're not comparing access key secret key and session token here?

remote_vector_index_builder/app/services/index_builder.py
navneet1v: We should use AWS_DEFAULT_REGION here.

remote_vector_index_builder/core/object_store/s3/s3_object_store.py
Rajrahane: if None is passed, which is the default, it follows the boto3 credential resoulution path

https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html
In this case it is actually passed as NULL
The boto3 client picks the ENV variables set starting with AWS_

remote_vector_index_builder/app/services/index_builder.py
jed326: If the correct region is required, then we should not have a default region at all right? It should be similar to
```
s3_endpoint_url = os.environ.get("S3_ENDPOINT_URL", None)
```

e2e/api/docker-compose.yml
Rajrahane: resolved, switched it to AWS_DEFAULT_REGION

remote_vector_index_builder/core/object_store/s3/s3_object_store_config.py
Rajrahane: added. missed it

remote_vector_index_builder/app/services/index_builder.py
Rajrahane: I should set it to None by default i think
So if you forget to set the AWS_DEFAULT_REGION env, it should break the client and not let it init right?

The S3ClientConfig pojo itself has it set to None
If you don't pass the region_name or pass a None, it fails to init the ObjectStore with a pydantic error (or more so the boto3 throws a connect error Forbidden 403 due to missing/incorrect region, when the pydantic validation was not present previously).

Rajrahane:
Important note I found: The @functools.cache (and @functools.lru_cache) decorators need __hash__ and __eq__ because they store function arguments as dictionary keys in their internal cache.
Hence the S3ClientConfig needs these functions implemented

remote_vector_index_builder/app/services/index_builder.py
Rajrahane: I had set it to us-west-2 because the initial implementation has a default
https://github.com/opensearch-project/remote-vector-index-builder/blob/main/remote_vector_index_builder/core/object_store/s3/s3_object_store.py#L134

If we don't set the AWS_REGION_NAME, the boto client fails to initialize (which is the expected behaviour)

Note the GH CI in knn will break instantly  I believe as it doesn't set it
 https://github.com/anntians/k-NN/blob/main/.github/workflows/remote_index_build.yml

Rajrahane: set the config such that if no region is found, boto3 client will fail to create since it defaults to None
effectively makes this a required param

Added docstrings in the required places

navneet1v (approved):

remote_vector_index_builder/app/services/index_builder.py
Rajrahane: Completed.
Resolving this conversation

PR #102, opened 2025-07-24
Change the app Dockerfile to use faiss-base as base image

### LINKED ISSUE #101: [BUG] API image CI does not use latest core image on github CI push to opensearch-staging
### What is the bug?
The github CI to publish the snapshot API to opensearch-staging does not use the latest core image. So if there is a change in the `core` library, then that change won't be reflected when building the `api` image. Reference: https://github.com/opensearch-project/remote-vector-index-builder/actions/runs/16426837535/job/46419159159. 

### How can one reproduce the bug?
By making a change in the `core` library, and then testing that the change isn't in the `api` image. 

### What is the expected behavior?
`api` image should always have latest changes from `core`, before CI push

rchitale7:
### Description
The `publish_remote_api_image` CI was failing, with this change: https://github.com/opensearch-project/remote-vector-index-builder/pull/98. The reason is that both `core` and `app` folders were modified. When publishing the `api` image, the GitHub CI used an older version of the `core` image, that did not have the latest modifications. This caused the CI to fail, because the code in the `app` folder is using the latest modifications to `core`. I fixed the app Dockerfile to follow the example from the `e2e` image: https://github.com/opensearch-project/remote-vector-index-builder/blob/main/e2e/api/Dockerfile, so that we're just copying the `app` and `core` directories onto the `faiss-base` image. This ensures any modifications to `core` will always be picked up by the `publish_remote_api_image` CI. 

I also updated some of the docs to reflect this change, along with the change from https://github.com/opensearch-project/remote-vector-index-builder/pull/98. 

### Issues Resolved
Resolves https://github.com/opensearch-project/remote-vector-index-builder/issues/101


navneet1v:
@rchitale7 I thought there is already separate CIs that looks at the what folder changes build the images. May be it can happen that core, app and base image cis are not running in order. I think there will be a way where we can add dependencies of CIs on one another so that they can run in order. Won’t that fix the problem?

rchitale7:
> @rchitale7 I thought there is already separate CIs that looks at the what folder changes build the images. May be it can happen that core, app and base image cis are not running in order. I think there will be a way where we can add dependencies of CIs on one another so that they can run in order. Won’t that fix the problem?

That does already exist here: https://github.com/opensearch-project/remote-vector-index-builder/blob/main/.github/workflows/publish_remote_core_image.yml#L67. So I'm not sure why this CI failed: https://github.com/opensearch-project/remote-vector-index-builder/actions/runs/16426837535/attempts/1 - maybe the GH runner erroneously used a cached image of `core`. I thought it made more sense to just copy the `core` and `app` folders instead of using `core` as a base image, since then we would never have this issue.

navneet1v:
> > @rchitale7 I thought there is already separate CIs that looks at the what folder changes build the images. May be it can happen that core, app and base image cis are not running in order. I think there will be a way where we can add dependencies of CIs on one another so that they can run in order. Won’t that fix the problem?
> 
> That does already exist here: https://github.com/opensearch-project/remote-vector-index-builder/blob/main/.github/workflows/publish_remote_core_image.yml#L67. So I'm not sure why this CI failed: https://github.com/opensearch-project/remote-vector-index-builder/actions/runs/16426837535/attempts/1 - maybe the GH runner erroneously used a cached image of `core`. I thought it made more sense to just copy the `core` and `app` folders instead of using `core` as a base image, since then we would never have this issue.

The problem with copying core only is we would now need to ensure that we install all the dependencies of core also + we setup the core. So basically whatever we were doing in core we need to do again. I think this could be a problem with github runners not running in order. If we can as API github runner to wait for core to complete and then ask core API runner to wait for base image CI to complete this problem will not happen.

PR #104, opened 2025-08-01
FP16 streaming support

0ctopus13prime:
### Description
This PR introduces streaming FP16 conversion when downloading FP32 vectors that uploaded by data node.
The logic presented in this PR is to have FP32ToFP16ConvertingBytesIO to take raw bytes, then directly convert the given bytes to FP16 vectors without additional allocation.

With this, we can guarantee that the maximum memory space would be bound to O(SizeOf(FP16 vectors)) even when FP32 vectors uploaded in S3.

### Issues Resolved
_List any issues this PR will resolve, e.g. Closes [...]._ 

### DIFF
--- a/remote_vector_index_builder/core/common/models/index_build_parameters.py
+++ b/remote_vector_index_builder/core/common/models/index_build_parameters.py
@@ -22,6 +22,7 @@ class DataType(str, Enum):
     """
 
     FLOAT = "float"
+    FLOAT16 = "half_float"
     BYTE = "byte"
     BINARY = "binary"
 
@@ -33,6 +34,8 @@ class DataType(str, Enum):
         """
         if self == DataType.FLOAT:
             return 4
+        elif self == DataType.FLOAT16:
+            return 2
         elif self == DataType.BYTE:
             return 1
         elif self == DataType.BINARY:
@@ -51,6 +54,7 @@ class SpaceType(str, Enum):
 
     L2 = "l2"
     INNERPRODUCT = "innerproduct"
+    HAMMING = "hamming"
 
 
 class Algorithm(str, Enum):
--- a/remote_vector_index_builder/core/fp32_to_fp16_converting_bytes_io.py
+++ b/remote_vector_index_builder/core/fp32_to_fp16_converting_bytes_io.py
@@ -1,0 +1,138 @@
+import os
+import numpy as np
+import threading
+
+"""
+This class converts raw FP32 byte streams into FP16 values and stores them in a pre-allocated FP16 NumPy array. It is
+designed to handle input arriving in multiple partitions from a continuous FP32 vector stream.
+Because a partition boundary may split an FP32 value across chunks, the class first identifies all complete FP32 values
+from the incoming bytes, converts them to FP16, and writes them into the FP16 array. Any remaining incomplete bytes are
+stored in a dictionary and reassembled once the rest of the bytes arrive in future partitions.
+
+For example, let's say we had 6 FP32 values as below:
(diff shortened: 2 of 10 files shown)


remote_vector_index_builder/core/fp32_to_fp16_converting_bytes_io.py
```diff
@@ -0,0 +1,99 @@
...
+        return True
+
+    def seek(self, offset, whence=0):
+        with self._lock:
+            if whence == os.SEEK_SET:
+                self._curr_offset = offset
+            elif whence == os.SEEK_CUR:
+                self._curr_offset += offset
+            elif whence == os.SEEK_END:
+                self._curr_offset = 4 * len(self._fp16_np)
```
rchitale7, line 23: nit: we can use a constant for 4 to show that it is 4 bytes for the float.

remote_vector_index_builder/core/fp32_to_fp16_converting_bytes_io.py
```diff
@@ -0,0 +1,99 @@
...
+            else:
+                raise ValueError(f"Unexpected whence={whence}")
+
+    def getbuffer(self):
+        return memoryview(self._fp16_np)
+
+    def write(self, b):
+        with self._lock:
```
rchitale7, line 31: i think a full example here would be helpful, its kind of hard to visualize this.

rchitale7 (approved):

remote_vector_index_builder/core/common/models/index_build_parameters.py
```diff
@@ -22,6 +22,7 @@ class DataType(str, Enum):
     """
 
     FLOAT = "float"
+    FLOAT16 = "float16"
```
navneet1v, line 25: I thought in the interface value will be half_float

navneet1v:
> This PR introduces streaming FP16 conversion when downloading FP32 vectors that uploaded by data node. The logic presented in this PR is to have FP32ToFP16ConvertingBytesIO to take raw bytes, then directly convert the given bytes to FP16 vectors without additional allocation.
> 
> With this, we can guarantee that the maximum memory space would be bound to O(SizeOf(FP16 vectors)) even when FP32 vectors uploaded in S3.

@0ctopus13prime what is the performance comparison we see this extra conversion is added? does this extra conversion take advantage to of SIMD?

0ctopus13prime:
> > This PR introduces streaming FP16 conversion when downloading FP32 vectors that uploaded by data node. The logic presented in this PR is to have FP32ToFP16ConvertingBytesIO to take raw bytes, then directly convert the given bytes to FP16 vectors without additional allocation.
> > With this, we can guarantee that the maximum memory space would be bound to O(SizeOf(FP16 vectors)) even when FP32 vectors uploaded in S3.
> 
> @0ctopus13prime what is the performance comparison we see this extra conversion is added? does this extra conversion take advantage to of SIMD?

Overall, the streaming approach shows 1.5 to 2× faster performance compared to downloading FP32 vectors first and then converting them to FP16. For example, when processing 1M FP32 vectors of 768 dimensions, the download-fp32-and-convert took 12,379 ms, whereas the streaming approach completed in 7,107 ms.

Both approaches rely on NumPy's type conversion and thus benefit from SIMD. However, the streaming avoids the memory reallocation overhead associated with dynamic buffer growth. When appending FP32 bytes to a BytesIO object during download, its internal buffer must grow, likely following a doubling reallocation strategy. In contrast, the streaming approach pre-allocates FP16 NumPy array upfront and writes directly into it during conversion. 

Therefore, streaming is expected 1.5 to 2x faster with half memory consumption.

0ctopus13prime:
@navneet1v @rchitale7 
Added comment with 'half_float' in the source.
Could you approve if it looks good on you? :)

navneet1v (approved):

PR #127, opened 2026-03-03
Adding skip_stored_vectors parameter to skip graph storage

rchitale7:
### Description
This PR introduces a `skip_stored_vectors` parameter to the remote vector index builder. When `skip_stored_vectors`: `True`, the remote vector index builder skips attaching the flat vector storage when converting the CAGRA graph to HNSW graph. This change allows the remote vector index builder to only upload the graph structure to the remote repository, and leaves it up to the k-NN plugin to stitch the graph and flat vectors back together at indexing or search time.There are three main benefits to this:

1. Make k-NN the owner of the flat vector storage. This makes it easier to make updates to the storage format in the future, without touching the RVIB
2. Reduce the memory pressure on the RVIB. Flat vectors can be very large, and take up almost 1/2 the total CPU memory required by the worker. 
3. Optimize the graph upload. Without the flat vectors, the graph upload is much faster

The k-NN plugin must send `skip_stored_vectors`: `True` in the build request, in order to trigger this behavior. For backwards compatibility with older versions of k-NN, this parameter defaults to `False` if not present. 

I opened an issue in faiss to try and get this changed merged upstream: https://github.com/facebookresearch/faiss/issues/4931 - ideally we have it merged upstream instead of relying on a patch.

To begin with, this change will be integrated with the quantization flow on k-NN. We are also doing something different here for quantization: instead of sending the quantized vectors from k-NN to RVIB, we send the fp32 vectors to RVIB. Then RVIB sends back the graph structure, and k-NN will stitch together the quantized vectors on the data node with the graph. The reason for this is to support 1 bit SQ quantization flat vector storage format in k-NN - `CAGRA` does not currently support this custom storage format, so we need to decouple the graph build with the flat vector storage format on the RVIB. This helps ensure k-NN owns the flat vector storage format, although it increase the memory pressure on the worker, and require longer download times. 

I will soon raise a PR on the k-NN side that supports stitching the graph back together with the storage, and link it here. 

### Testing

Unit tests. 

I also did a manual test where i spun up a local instance of the API docker container, and triggered a remote build request with the following script:

```
curl -XPOST "http://0.0.0.0:80/_build" \
-H 'Content-Type: application/json' \
-d '
    {
        "repository_type": "s3",
        "container_name": "testbucket-rchital",
        "vector_path": "test-data/more-data/1536_2000000.knnvec",
        "doc_id_path": "test-data/more-data/1536_2000000_ids.knndid",
        "dimension": "1536",
        "doc_count": "2000000",
        "skip_stored_vectors": true
    }
'
```

This script uses a 2 mil x 1536 random vector dataset in my local s3 bucket. I verified the build was successful, also monitored the CPU memory using the following script: https://github.com/opensearch-project/remote-vector-index-builder/blob/main/scripts/monitor_memory.py . The peak CPU memory was 13 GB - well below the 25 GB peak usage observed when `skip_stored_vectors`: `false`.

I also tested without the `skip_stored_vectors` flag to test for backwards compatibility. 

```
curl -XPOST "http://0.0.0.0:80/_build" \
-H 'Content-Type: application/json' \
-d '
    {
        "repository_type": "s3",
        "container_name": "testbucket-rchital",
        "vector_path": "test-data/more-data/1536_2000000.knnvec",
        "doc_id_path": "test-data/more-data/1536_2000000_ids.knndid",
        "dimension": "1536",
        "doc_count": "2000000"
    }
'
```

I verified this is continuing to work as expected. The integration test CIs also don't use this flag, and are passing, so we can be confident the addition of this new parameter to the API won't break backwards compatibility. 

### Issues Resolved
Partially for: https://github.com/opensearch-project/k-NN/issues/3274

### DIFF
--- a/API.md
+++ b/API.md
@@ -16,7 +16,7 @@ Request Parameters
     "tenant_id": "UniqueClusterID", // Unique identifier for the cluster making the request
     "dimension": 768,
     "doc_count": 1000000, // Must be greater than 4
-    "data_type": "float",
+    "data_type": // {float, half_float, byte, binary}
     "engine": "faiss",
     "index_parameters": {
         "space_type": // {l2, innerproduct}
@@ -26,7 +26,8 @@ Request Parameters
             "ef_search": 100,
             "m": 16
         }
-    }
+    }, 
+    "skip_stored_vectors": false
 }
 
 Request Response:
@@ -47,6 +48,7 @@ Request Response:
 * By including an `algorithm` field in index parameters, we leave the door open for IVF, future algorithms
 * The existing algorithm parameters will be type checked on the client and server side, but the size of the map is variable to allow for future parameter additions
 * Qualitative parameters like `repository_type`, `data_type`, `engine`, `algorithm`, and `space_type` currently only support the options listed. The other numerical and string settings follow k-NN/repository snapshot precedent on ranges and expected values: https://opensearch.org/docs/latest/search-plugins/knn/knn-index/
+* `skip_stored_vectors` controls whether to send only the HNSW graph structure back, or the HNSW graph structure with flat vectors attached. It defaults to `false`. If `true`, the k-NN data node must handle stitching the graph back together with the flat vectors at indexing or search time. 
 
 #### Error codes
 
--- a/base_image/build_scripts/Dockerfile
+++ b/base_image/build_scripts/Dockerfile
@@ -18,6 +18,8 @@ RUN chown -R appuser:appuser /opt
 
 RUN mkdir /tmp/faiss
 COPY ./base_image/faiss /tmp/faiss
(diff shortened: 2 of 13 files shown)


API.md
```diff
@@ -26,7 +26,8 @@ Request Parameters
             "ef_search": 100,
             "m": 16
         }
-    }
+    }, 
+    "graph_only": False
```
owenhalpert: nit: json uses `false`

owenhalpert (dismissed):

API.md
navneet1v: we should name this better. May be something like skip_stored_vectors

owenhalpert (approved):

navneet1v (approved):

